PR #1, opened 2022-02-07
Add game state interface

atilbian:


src/gamestate.ts
```diff
@@ -0,0 +1,12 @@
+interface GameState {
+  board: Array<Array<Cell>>;
+}
+
+interface Cell {
+  value: CellValue;
+}
+
+enum CellValue {
```
fzaninotto, line 9: I recommend that you don't use enums in typeScript https://www.executeprogram.com/blog/typescript-features-to-avoid

slax57: We'll merge it as is, to unlock future devs, and we'll change this later. Thanks for pointing this out 🙂

fzaninotto: use union type instead

```tsx
type CellValue = 'emtpy' | 'black';
```

PR #2, opened 2022-02-07
Feature 1 - Render a game state

atilbian:
- [x] Draw a game from a GameState object
- [x] Print the rendered game in the console
- [x] Test it
- [x] Make the expected game more readable in UT

### DIFF
--- a/src/gamestate.ts
+++ b/src/gamestate.ts
@@ -1,12 +1,12 @@
-interface GameState {
+export interface GameState {
   board: Array<Array<Cell>>;
 }
 
-interface Cell {
+export interface Cell {
   value: CellValue;
 }
 
-enum CellValue {
+export enum CellValue {
   Empty,
   Black,
 }
--- a/src/render.spec.ts
+++ b/src/render.spec.ts
@@ -14,9 +14,10 @@ describe("Render a boad in string from a game state object", () => {
     const renderedGameState = getRenderedGameState(twoRowsAllCellsEmpty);
 
     expect(renderedGameState).toBe(
-      `⬡⬡
+      `
+⬡⬡
  ⬡⬡
-`
+`.substring(1)
     );
   });
   it("Should get a 2-rows board with all cells black", () => {
@@ -30,9 +31,10 @@ describe("Render a boad in string from a game state object", () => {
     const renderedGameState = getRenderedGameState(twoRowsWithAllCellsBlack);
 
     expect(renderedGameState).toBe(
-      `⬢⬢
+      `
+⬢⬢
  ⬢⬢
-`
+`.substring(1)
     );
   });
   it("Should get a 2-rows board with mixed cells values", () => {
@@ -46,9 +48,10 @@ describe("Render a boad in string from a game state object", () => {
     const renderedGameState = getRenderedGameState(twoRowsWithMixedCellValues);
 
     expect(renderedGameState).toBe(
-      `⬡⬢
+      `
+⬡⬢
  ⬡⬢
-`
+`.substring(1)
     );
   });
   it("Should get a 5-rows board with mixed cells values", () => {
@@ -65,12 +68,13 @@ describe("Render a boad in string from a game state object", () => {
     const renderedGameState = getRenderedGameState(fiveRowsWithMixedCellValues);
 
     expect(renderedGameState).toBe(
-      `⬡⬢
+      `
+⬡⬢
  ⬡⬢
   ⬡⬢
    ⬡⬢
     ⬡⬢
-`
+`.substring(1)
     );
   });
   it("Should get an empty board", () => {
@@ -93,9 +97,10 @@ describe("Render a boad in string from a game state object", () => {
     const renderedGameState = getRenderedGameState(nonSquareBoard);
 
     expect(renderedGameState).toBe(
-      `⬡
+      `
+⬡
  ⬡⬢
-`
+`.substring(1)
     );
   });
 });
--- a/src/render.spec.ts
+++ b/src/render.spec.ts
@@ -2,35 +2,56 @@ import expect from "expect";
 import { getRenderedGameState } from "./render";
 import { GameState, CellValue } from "./gamestate";
 
-describe("Draw a boad from a game state object", () => {
-  it("Draw 2-rows board with all cells empty", () => {
+describe("Render a boad in string from a game state object", () => {
+  it("Should get a 2-rows board with all cells empty", () => {
     const twoRowsAllCellsEmpty: GameState = {
       board: [
         [{ value: CellValue.Empty }, { value: CellValue.Empty }],
         [{ value: CellValue.Empty }, { value: CellValue.Empty }],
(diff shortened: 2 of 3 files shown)


src/render.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    board: [
+      [{ value: CellValue.Empty }, { value: CellValue.Black }],
+      [{ value: CellValue.Empty }, { value: CellValue.Black }],
+    ],
+  };
+}
+
+describe("Get 2-rows board with all cells empty must return it", () => {
+  it("should pass", () => {
+    expect(getRenderedGameState(get2RowsGameStateAllCellsEmpty())).toBe(
```
fzaninotto: not readable: I need to move up and down to look at the test and the input.

src/render.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+      "⬡⬡\n ⬡⬡"
```
fzaninotto: Not readable. Please use backticks to that the expected outputs looks like an actual board

src/render.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+function get2RowsGameStateMixedCellsValues(): GameState {
+  return {
+    board: [
+      [{ value: CellValue.Empty }, { value: CellValue.Black }],
+      [{ value: CellValue.Empty }, { value: CellValue.Black }],
+    ],
+  };
+}
+
+describe("Get 2-rows board with all cells empty must return it", () => {
```
fzaninotto: We usually wite one "describe" per main function, and many "it" inside for each case

src/render.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    );
+  });
+});
+
+describe("Get 2-rows board with mixed cells values must return it", () => {
+  it("should pass", () => {
+    expect(getRenderedGameState(get2RowsGameStateMixedCellsValues())).toBe(
+      "⬡⬢\n ⬡⬢"
+    );
+  });
```
fzaninotto, line 105: you're probbaly missing a test with a bigger board.

fzaninotto (changes requested):

src/render.ts
```diff
@@ -0,0 +1,26 @@
...
+    const initialSpace = spaceCharacter.repeat(index);
+    let renderedRow = initialSpace;
+    row.forEach((cell) => {
+      const renderedCell = cell.value == CellValue.Black ? "⬢" : "⬡";
+      renderedRow += renderedCell;
+    });
+    if (index > 0) {
+      result += lineSeparatorCharacter;
+    }
+    result += renderedRow;
```
slax57:
```suggestion
    result += renderedRow;
    result += lineSeparatorCharacter;
```

src/render.spec.ts
atilbian: I'm not sure to understand how to do it with backticks. Could we check that tomorrow?

src/gamestate.ts
```diff
@@ -1,12 +1,12 @@
...
   board: Array<Array<Cell>>;
 }
 
-interface Cell {
+export interface Cell {
   value: CellValue;
 }
 
-enum CellValue {
+export enum CellValue {
```
mchaffotte, line 9: use union instead of enum

src/render.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+describe("Draw a boad from a game state object", () => {
+  it("Draw 2-rows board with all cells empty", () => {
+    const twoRowsAllCellsEmpty: GameState = {
+      board: [
+        [{ value: CellValue.Empty }, { value: CellValue.Empty }],
+        [{ value: CellValue.Empty }, { value: CellValue.Empty }],
+      ],
+    };
+    expect(getRenderedGameState(twoRowsAllCellsEmpty)).toBe("⬡⬡\n ⬡⬡\n");
```
mchaffotte: Use the Given/When/The pattern in order to ease the test readability

src/render.spec.ts
```diff
@@ -0,0 +1,47 @@
+import expect from "expect";
+import { getRenderedGameState } from "./render";
+import { GameState, CellValue } from "./gamestate";
+
+describe("Draw a boad from a game state object", () => {
+  it("Draw 2-rows board with all cells empty", () => {
```
mchaffotte: it("should draw .....", ....

src/render.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+        [{ value: CellValue.Empty }, { value: CellValue.Black }],
+        [{ value: CellValue.Empty }, { value: CellValue.Black }],
+        [{ value: CellValue.Empty }, { value: CellValue.Black }],
+      ],
+    };
+    expect(getRenderedGameState(fiveRowsWithMixedCellValues)).toBe(
+      "⬡⬢\n ⬡⬢\n  ⬡⬢\n   ⬡⬢\n    ⬡⬢\n"
+    );
+  });
+});
```
mchaffotte, line 106: You should think about tests using limit cases / exceptions (for example: with an empty array)

src/render.ts
```diff
@@ -0,0 +1,24 @@
+import { GameState, CellValue } from "./gamestate";
+
+export function displayGameState(gameState: GameState) {
+  const renderedGameState = getRenderedGameState(gameState);
+  console.log(renderedGameState);
+}
+
+export function getRenderedGameState(gameState: GameState): string {
+  const spaceCharacter = " ";
+  const lineSeparatorCharacter = "\n";
```
mchaffotte: extract constants outside of the function (-> each time the function is called these variables are built again)

src/render.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  let result = "";
+  gameState.board.forEach((row, index) => {
+    const initialSpace = spaceCharacter.repeat(index);
+    let renderedRow = initialSpace;
+    row.forEach((cell) => {
+      const renderedCell = cell.value == CellValue.Black ? "⬢" : "⬡";
```
mchaffotte: extract your pawns in contants

mchaffotte: It should be better to extract the inner loop for a better readability

mchaffotte (changes requested):

src/render.ts
slax57: names should be uppercase?

src/gamestate.ts
slax57: will be resolved with https://github.com/marmelab/Hex/pull/3#discussion_r801038301

src/render.spec.ts
atilbian: Am i using it correctly ?

```
const emptyBoard = {
      board: [],
};

const renderedGameState = getRenderedGameState(emptyBoard);

expect(renderedGameState).toBe("");
```

src/render.ts
mchaffotte: Yes it must be ;)

PR #3, opened 2022-02-07
Feature 1 - Implements config file parsing

slax57:
- [x] Implement `parseConfigFile()` function
- [x] Rebase once https://github.com/marmelab/Hex/pull/2 is merged
- [x] Create unit tests for this function

## Demo

2022-02-08 14_24_23-configfile json - Hex  WSL_ Ubuntu-20 04  - Visual Studio Code


.vscode/launch.json
```diff
@@ -0,0 +1,22 @@
+{
```
fzaninotto: This is a developer-specific file. Please don't commit this file (add it to .gitignore).

src/parseConfigFile.ts
```diff
@@ -0,0 +1,47 @@
+import * as fs from "fs";
+import { Cell, CellValue, GameState } from "./gamestate";
+
+export function parseConfigFile(filePath: string): GameState {
```
fzaninotto: you have 3 "parse" functions, but one loads a file, and the 2 others parse a string. Perhaps you should split the first one so that you have one load ans one parse function. Besides, it would make it easier to test.

fzaninotto (changes requested):

.vscode/launch.json
slax57: Removed the file from control version

src/parseConfigFile.ts
slax57: Made another attempt, please notify me if I still got it wrong 🙂

src/gamestate.ts
```diff
@@ -1,12 +1,12 @@
...
   board: Array<Array<Cell>>;
 }
 
-interface Cell {
+export interface Cell {
   value: CellValue;
 }
 
-enum CellValue {
+export enum CellValue {
```
mchaffotte: use union instead of enum

src/index.spec.ts
```diff
@@ -1,7 +1,16 @@
...
+describe("load game state from a valid config file", () => {
+  it("should load configfile01.json without error", () => {
+    const expected: GameState = {
+      board: [
+        [ { value: CellValue.Black }, { value: CellValue.Empty }, { value: CellValue.Black } ],
+        [ { value: CellValue.Empty }, { value: CellValue.Empty }, { value: CellValue.Empty } ],
+        [ { value: CellValue.Empty }, { value: CellValue.Empty }, { value: CellValue.Empty } ]
+      ]
+    };
+    expect(loadGameStateFromFile("src/fixtures/configfile01.json")).toStrictEqual(expected);
```
mchaffotte: use the Given/When/Then pattern to ease the test readability

src/index.spec.ts
```diff
@@ -1,7 +1,16 @@
...
   });
```
mchaffotte: Think about tests using limit cases and exceptions

src/parseConfigFile.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+function parseCellValue(cell: any): CellValue {
+  if (cell.value == null) {
+    return CellValue.Empty;
+  } else if (cell.value == "1") {
+    return CellValue.Black;
+  } else {
+    throw Error(`Cannot parse config file: Invalid value: ${cell.value}`);
+  }
```
mchaffotte: The elses are useless in this function as your use a return statement

src/parseConfigFile.ts
```diff
@@ -0,0 +1,51 @@
...
+    let parsedCell = { value: parseCellValue(cell) };
+    parsedRow.push(parsedCell);
+  }
+  return parsedRow;
+}
+
+function parseCellValue(cell: any): CellValue {
+  if (cell.value == null) {
+    return CellValue.Empty;
+  } else if (cell.value == "1") {
```
mchaffotte: 1 is a magic number

src/parseConfigFile.ts
```diff
@@ -0,0 +1,51 @@
...
+    throw Error(`Malformed config file ${filePath}`);
+  }
+  if (row.length !== rowsCount) {
+    throw Error(
+      `Malformed config file ${filePath}: all rows should contain ${rowsCount} cells`
+    );
+  }
+  let parsedRow: Array<Cell> = [];
+  for (let cell of row) {
+    let parsedCell = { value: parseCellValue(cell) };
```
mchaffotte: use const instead of let whenever it is possible

mchaffotte: why the parseCellValue returns only a value and not a Cell?

src/parseConfigFile.ts
```diff
@@ -0,0 +1,51 @@
...
+function parseConfigFile(filePath: string): any {
+  try {
+    return JSON.parse(fs.readFileSync(filePath, "utf-8"));
+  } catch (error) {
+    console.error(error);
+    throw Error(`Unable to parse config file ${filePath}`);
+  }
+}
+
+function parseRow(row: any, rowsCount: number, filePath: string): Array<Cell> {
```
mchaffotte, line 43: row should be typed as you iterate over it in the function

src/parseConfigFile.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export function loadGameStateFromFile(filePath: string): GameState {
+  let data = parseConfigFile(filePath);
+  if (!Array.isArray(data)) {
+    throw Error(`Malformed config file ${filePath}`);
+  }
+  let rowsCount = data.length;
+  let gameState = { board: [] };
+  for (let row of data) {
+    gameState.board.push(parseRow(row, rowsCount, filePath));
```
mchaffotte: You break the Demeter law. You should build your board and finally build your gameState

src/index.spec.ts
slax57: Done. Hope it is better 🙂

src/index.spec.ts
slax57: Added test cases for :
* Invalid JSON File
* Wrong number of columns in a row
* Invalid value (like 2 for instance)

src/parseConfigFile.ts
slax57: Reworked this function, this comment is no more applicable

src/parseConfigFile.ts
slax57: moved it to a constant

src/parseConfigFile.ts
slax57: used `const` when possible and reworked `parseCellValue` function so that it returns a Cell

src/parseConfigFile.ts
slax57: fixed

src/parseConfigFile.ts
slax57: Indeed I could have used `JSON.parse(fileContent) as Array<Array<Cell>>` or stg close, but we still need to map each cell to a `Cell` object, which would have to be done manually.

Keeping actual impl for now

src/parseConfigFile.ts
```diff
@@ -0,0 +1,72 @@
...
+    parsedRow.push(parseCell(cell));
+  }
+  return parsedRow;
+}
+
+function parseCell(cell: any): Cell {
+  let parsedCellValue = CellValue.Empty;
+  switch (cell.value) {
+    case null:
+      break;
```
mchaffotte: case null do nothing! It means that you miss something. You should instantiate parsedCellValue inside your switch case

src/index.ts
```diff
@@ -1 +1,27 @@
...
+});
+
+// Check mandatory parameters
+if (!filePath) {
+  console.error("No config file was provided");
+  console.error("usage: yarn start -f=myconfigfile");
+  throw Error("No config file was provided");
+}
+
+// Parse config file into GameState object
```
atilbian: Seems like it's too redundant with the related code

src/parseConfigFile.ts
```diff
@@ -0,0 +1,72 @@
...
+    case BLACK_PAWN_VALUE:
+      parsedCellValue = CellValue.Black;
+      break;
+    default:
+      throw Error(`Cannot parse config file: Invalid value: ${cell.value}`);
+  }
+  const parsedCell: Cell = { value: parsedCellValue };
+  return parsedCell;
```
mchaffotte: Do not need to create a local variable for nothing

mchaffotte (changes requested):

src/index.ts
```diff
@@ -1 +1,27 @@
...
+let gameState = parseGameStateFromFile(filePath);
+
+console.log(gameState);
```
atilbian: Need to be replaced by the call to the drawing function

slax57: indeed!! 😅

PR #4, opened 2022-02-08
Feature 2 allow player to add black stones

atilbian:
- [x] Prompt message to user / parse his targeted cell
- [x] Update game state based on user input
- [x] Create a function to map from user coordinates to our data struct coordinates
- [x] Create the game loop
- [x] Add ut
- [ ] ~Update the render function to display coordinates~
- [x] Init the game state from a file or create a new one if no file is given

## Demo
demo-feature-2

Related to https://trello.com/c/oF19QqKC/2-1-as-norbert-i-want-to-be-able-to-add-a-stone-anywhere-on-the-board


fzaninotto (commented):
Nice! You probably need to split the game scrim though, because it's becoming too big. Also, unit tests.

src/game.ts
```diff
@@ -0,0 +1,147 @@
...
+    );
+  }
+  if (/[A-Z]/.test(userCoordinates[0])) {
+    x = userCoordinates[0].charCodeAt(0) - UTF16_CODE_OF_LETTER_A;
+  } else {
+    throw new Error(`Given character is not included in the range [A,Z]`);
+  }
+  y = parseInt(userCoordinates.slice(1, userCoordinates.length));
+  if (y >= 0 && y <= 25) {
+    return { x: x, y: y };
```
mchaffotte: you can return straightly {x, y}

mchaffotte: Is it required all of these errors? If not I will be easier to check only a regexp and then get the value of x an y.

src/game.ts
```diff
@@ -0,0 +1,147 @@
...
+  } else {
+    throw new Error(`Given number is not included in the range of [0,25]`);
+  }
+}
+
+export function generateNewBoard(): GameState {
```
mchaffotte: There are too many responsibilities (gameloop/init/...) in this file. You should split them in several files.

src/game.ts
```diff
@@ -0,0 +1,147 @@
...
+// console.log(mapFromUserInputToBoardCoordinates("C2"));
+
+// console.log(mapFromUserInputToBoardCoordinates("e2"));
+// console.log(mapFromUserInputToBoardCoordinates("eee2"));
+// console.log(mapFromUserInputToBoardCoordinates("E200"));
+
+// console.log(isCoordinatesInBoard(generateNewBoard(), 1, 1));
+// console.log(isCoordinatesInBoard(generateNewBoard(), 4, 1));
+// console.log(isCoordinatesInBoard(generateNewBoard(), 0, 4));
+// console.log(isCoordinatesInBoard(generateNewBoard(), 4, 0));
```
mchaffotte: :scream:

src/game.ts
```diff
@@ -0,0 +1,147 @@
...
+  if (coordinates.y < 0 || coordinates.y >= gameState.board.length) {
+    return false;
+  }
+  if (
+    coordinates.x < 0 ||
+    coordinates.x >= gameState.board[coordinates.y].length
+  ) {
+    return false;
+  }
+  return true;
```
mchaffotte: You can ease this function according to the name of this function. You are checking that the coordinates are valid so check only that return y inside the board && x inside the board.

mchaffotte (changes requested):

src/game.ts
```diff
@@ -0,0 +1,39 @@
...
+        const boardCoordinates = getBoardCoordinatesFromUserInput(userInput);
+        currentState = updateGameState(currentState, boardCoordinates);
+        areUserCoordinatesValid = true;
+      } catch (error) {
+        console.error(`${error.message} Please try again.`);
+      }
+    }
+    displayGameState(currentState);
+  }
+}
```
slax57, line 26: rework suggestion:

```suggestion
export async function runGameLoop(initialGameState: GameState) {
  let currentState = initialGameState;
  let continueGame = true;
  while (continueGame) {
    displayGameState(currentState);
    let areUserCoordinatesValid = false;
    while (!areUserCoordinatesValid) {
      try {
        const userInput = await askCellCoordinatesToUser();
        const boardCoordinates = getBoardCoordinatesFromUserInput(userInput);
        currentState = updateGameState(currentState, boardCoordinates);
        areUserCoordinatesValid = true;
      } catch (error) {
        console.error(`${error.message} Please try again.`);
      }
    }
  }
}
```

src/gamestate.ts
```diff
@@ -1,7 +1,59 @@
...
+      ],
+      [
+        { value: "empty" },
+        { value: "empty" },
+        { value: "empty" },
+        { value: "empty" },
+      ],
+    ],
+  };
+}
```
slax57, line 85: Suggestion (untested):

```suggestion
export function generateNewBoard(n: number): GameState {
  return { board: Array(n).fill(Array(n).fill({ value: "empty" })) };
}
```

src/gamestate.ts
```diff
@@ -1,7 +1,59 @@
...
+
+export function updateGameState(
+  previousState: GameState,
+  actionCoordinates: Coordinates
+): GameState {
```
slax57: Imo `nextMove` would be a better name for `actionCoordinates`, what do you think?

src/gamestate.ts
```diff
@@ -1,7 +1,59 @@
...
+  if (!areCoordinatesValid(previousState, actionCoordinates)) {
```
slax57: Since `areCoordinatesValid()` uses the game state and not only the user input, I feel like it would rather need to be declared in this file (`gamestate.ts`) rather than in `user.ts`. What do you think?

src/gamestate.ts
```diff
@@ -1,7 +1,59 @@
...
+    throw new Error("Given coordinates are outside the scope of the board.");
+  }
+  if (
+    previousState.board[actionCoordinates.y][actionCoordinates.x].value !==
+    "empty"
```
slax57: Those lines could be part of `areCoordinatesValid()` according to me

src/user.ts
```diff
@@ -0,0 +1,57 @@
+import { createInterface } from "readline";
+import { GameState } from "./gamestate";
+import { UTF16_CODE_OF_LETTER_A, Coordinates } from "./utils";
+
+export function getBoardCoordinatesFromUserInput(
+  userCoordinates: string
+): Coordinates {
+  let x: number, y: number;
+  if (userCoordinates.length < 2) {
```
slax57: What if the board is larger than 9x9 ? Stg like "K11" will be rejected although it's valid 😬

src/user.ts
```diff
@@ -0,0 +1,57 @@
...
+      `Given coordinates is not compatible. Must be something like "H4"`
+    );
+  }
+  if (/[A-Z]/.test(userCoordinates[0])) {
+    x = userCoordinates[0].charCodeAt(0) - UTF16_CODE_OF_LETTER_A;
+  } else {
+    throw new Error(`Given character is not included in the range [A,Z]`);
+  }
+  y = parseInt(userCoordinates.slice(1, userCoordinates.length));
+  if (y >= 0 && y <= 25) {
```
slax57: Wouldn't it be more user-friendly to have row numbers starting at 1?

slax57 (changes requested):

slax57:
Potential conflicts/uniformisation to handle after https://github.com/marmelab/Hex/pull/5 has been merged:

- [x] `Coordinates` interface in `utils.ts`
- [x] `doesCellExist()` function in `gamestate.ts`

PR #5, opened 2022-02-09
Feature 3 implement win detection

slax57:
## Content

Make use of the library https://github.com/bbecquet/jkstra to implement path finding to determine when the user has won the game

## Checklist

- [x] Merge `Coordinates` interface with the one created in https://github.com/marmelab/Hex/pull/4 -> moved here: https://github.com/marmelab/Hex/pull/4#issuecomment-1033942046
- [x] Reorganize source files: move common interfaces to a specific file

## Demo

hex-feature-3


src/pathfinding.ts
```diff
@@ -0,0 +1,173 @@
...
+  const path_black = dijkstra.shortestPath(
+    hexBoardGraph.vertexMap.get("black-start"),
+    hexBoardGraph.vertexMap.get("black-end")
+  );
+  const path_white = dijkstra.shortestPath(
+    hexBoardGraph.vertexMap.get("white-start"),
+    hexBoardGraph.vertexMap.get("white-end")
+  );
+
+  return !!(path_black || path_white);
```
atilbian: Is their a reason to use two not operator ? Seems like we could omit them.

src/pathfinding.ts
```diff
@@ -0,0 +1,173 @@
...
+  vertex2: string
+) {
+  hexBoardGraph.graph.addEdgePair(
+    hexBoardGraph.vertexMap.get(vertex1),
+    hexBoardGraph.vertexMap.get(vertex2),
+    1
+  );
+}
+
+function existsPathForPlayer(hexBoardGraph: HexBoardGraph): boolean {
```
atilbian: Could be renamed to be more explicit about the fact it return a bool ?

src/pathfinding.ts
```diff
@@ -0,0 +1,173 @@
...
+  stone_color: "black"
+): boolean {
+  if (cell.y < 0 || cell.y >= gameState.board.length) {
+    return false;
+  }
+  if (cell.x < 0 || cell.x >= gameState.board[cell.y].length) {
+    return false;
+  }
+  return gameState.board[cell.y][cell.x].value == stone_color;
+}
```
atilbian: It could be splitted into 2 differents func(one for checking if the cell exist and other to check if the cell already have a pawn on it) to be more reusable. And it's not really related to the implem of the graph/pathfinding so it could be extracted into another module.

src/pathfinding.ts
```diff
@@ -0,0 +1,173 @@
...
+    addEdge(hexBoardGraph, `${gameState.board.length - 1}-${i}`, "white-end");
+  });
+}
+
+function createEdgePairForNeighbor(
+  gameState: GameState,
+  hexBoardGraph: HexBoardGraph,
+  stone_color: "black",
+  current_cell: Coordinates,
+  neighbor: Coordinates
```
atilbian: Mixed used of naming conventions

src/pathfinding.ts
```diff
@@ -0,0 +1,173 @@
...
+      addVertex(hexBoardGraph, `${y}-${x}`);
+    });
+  });
+}
+
+function createEdgesFromGameState(
+  gameState: GameState,
+  hexBoardGraph: HexBoardGraph
+) {
+  const stone_color = "black";
```
mchaffotte: your variable is written in camelCase not in snakeCase -> this constant should be extracted.

src/pathfinding.ts
```diff
@@ -0,0 +1,173 @@
...
+  y: number;
+}
+
+export function playerHasWon(gameState: GameState): boolean {
+  const hexBoardGraph = createGraphFromGameState(gameState);
+  return existsPathForPlayer(hexBoardGraph);
+}
+
+function createGraphFromGameState(gameState: GameState): HexBoardGraph {
+  let hexBoardGraph: HexBoardGraph = {
```
mchaffotte: use const instead of let

src/pathfinding.ts
atilbian: Seems lile this const could be extracted at the top of the file

src/pathfinding.ts
```diff
@@ -0,0 +1,173 @@
...
+  hexBoardGraph: HexBoardGraph
+) {
+  const stone_color = "black";
+
+  // Add edge pair for all nodes with one of its 3 possible neighbors
+  // if the stone color matches
+  gameState.board.forEach((row, y) => {
+    row.forEach((cell, x) => {
+      if (cell.value == stone_color) {
+        const current_cell: Coordinates = { y: y, x: x };
```
mchaffotte: naming

src/pathfinding.ts
```diff
@@ -0,0 +1,173 @@
...
+    addEdge(hexBoardGraph, "white-start", `0-${i}`);
+    addEdge(hexBoardGraph, `${gameState.board.length - 1}-${i}`, "white-end");
+  });
+}
+
+function createEdgePairForNeighbor(
+  gameState: GameState,
+  hexBoardGraph: HexBoardGraph,
+  stone_color: "black",
+  current_cell: Coordinates,
```
mchaffotte: naming

src/pathfinding.ts
```diff
@@ -0,0 +1,173 @@
...
+      `${current_cell.y}-${current_cell.x}`,
+      `${neighbor.y}-${neighbor.x}`
+    );
+  }
+}
+
+function cellExistsAndHasStone(
+  gameState: GameState,
+  cell: Coordinates,
+  stone_color: "black"
```
mchaffotte: naming

src/pathfinding.ts
```diff
@@ -0,0 +1,173 @@
...
+  vertex2: string
+) {
+  hexBoardGraph.graph.addEdge(
+    hexBoardGraph.vertexMap.get(vertex1),
+    hexBoardGraph.vertexMap.get(vertex2),
+    1
+  );
+}
+
+function addEdgePair(
```
mchaffotte: what this function does? edgePair is not clear enough

src/pathfinding.ts
```diff
@@ -0,0 +1,173 @@
...
+
+interface Coordinates {
+  x: number;
+  y: number;
+}
+
+export function playerHasWon(gameState: GameState): boolean {
+  const hexBoardGraph = createGraphFromGameState(gameState);
+  return existsPathForPlayer(hexBoardGraph);
+}
```
atilbian: I feel like it's could be extracted from this pathfinding module to avoid mixing technical implementation and the game rules

mchaffotte: It think that is not in the right file. There are too many responsibilities in this file -> create the graph and path finding and a game rule. You should separate this logic.

mchaffotte (changes requested):

src/pathfinding.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+                [{ value: "black" }, { value: "empty" }, { value: "empty" }],
+            ],
+        };
+        // when
+        const ouput = playerHasWon(input);
+        // then
+        expect(ouput).toStrictEqual(true);
+    });
+
+    it("should NOT say WIN with a 3x3 board with an incomplete vertical path", () => {
```
fzaninotto: this is misleading, because the array of arrays looks like an alignment. I know it is not aligned with hexagons, but this means the tests aren't easy to maintain. 

Perhaps you need a utility function for tests, which returns a game state from a multiline string.

```jsx
convertStringToState(`
⬡ ⬢ ⬡
 ⬡ ⬢ ⬡
  ⬡ ⬡ ⬡
`);
// [ 
//    [{ value: "empty" }, { value: "black" }, { value: "empty" }],
//    [{ value: "empty" }, { value: "black" }, { value: "empty" }],
//    [{ value: "empty" }, { value: "empty" }, { value: "empty" }],
// ]
```

src/pathfinding.ts
slax57: Actually it's a choice to have it inside the function and not as a constant, since later it will have to be passed as one of the functions arguments.

Will fix the case though.

src/pathfinding.ts
slax57: Yes, the `!!` operator ensures I get a boolean as the function output. I had to use this trick, because the jkstra lib returns `null` if no path is found, and apparenlty `null || null` equals `null` in JS, which is not what I want

src/pathfinding.ts
slax57: see https://github.com/marmelab/Hex/pull/5#discussion_r802744038

src/pathfinding.ts
slax57: split this file to:
- graph.ts
- pathfinding.ts
- gamestate.ts

src/pathfinding.ts
slax57: split into `doesCellExist()` and `doesCellHaveStone()` and moved to `gamestate.ts`

src/pathfinding.ts
slax57: Indeed, same comment as https://github.com/marmelab/Hex/pull/5#discussion_r802689959

Taken into account 😉

PR #7, opened 2022-02-10
Feature add white player

atilbian:
- [x] Render a white stone
- [x] Switch between white/black player in game loop and tell who is the next to play
- [x] Allow players to put a stone of their related color
- [x] Stop game loop when a player has won and congragulate him
- [x] Update and ~add tests~
- [x] Allow to import game state from file with white stones in it

Related to : https://trello.com/c/SwXWtQyu/4-1-as-norbert-i-want-to-be-able-to-play-against-someone-else

## Demo
demo-feature-add-white-player

### DIFF
--- a/src/game.ts
+++ b/src/game.ts
@@ -2,7 +2,7 @@ import {
   GameState,
   updateGameState,
   playerHasWon,
-  stoneColor,
+  StoneColor,
 } from "./gameState";
 import {
   askCellCoordinatesToUser,
@@ -15,7 +15,7 @@ export async function runGameLoop(initialGameState: GameState) {
   let currentState = initialGameState;
   let hasBlackWon = playerHasWon(currentState, "black");
   let hasWhiteWon = playerHasWon(currentState, "white");
-  let activeStoneColor: stoneColor = "white";
+  let activeStoneColor: StoneColor = "white";
   while (!hasBlackWon && !hasWhiteWon) {
     let areUserCoordinatesValid = false;
     while (!areUserCoordinatesValid) {
--- a/src/game.ts
+++ b/src/game.ts
@@ -1,4 +1,9 @@
-import { GameState, updateGameState, playerHasWon } from "./gameState";
+import {
+  GameState,
+  updateGameState,
+  playerHasWon,
+  stoneColor,
+} from "./gameState";
 import {
   askCellCoordinatesToUser,
   getBoardCoordinatesFromUserInput,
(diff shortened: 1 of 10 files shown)


src/game.ts
```diff
@@ -23,4 +23,5 @@ export async function runGameLoop(initialGameState: GameState) {
     }
     displayGameState(currentState);
   }
+  console.log("Player has won the game!");
```
fzaninotto: which player?

src/gameState.spec.ts
```diff
@@ -150,24 +150,99 @@ describe("Init a new board from scratch", () => {
           { value: "empty" },
```
fzaninotto: save yourself some copy pasting:

```js
Array(9).fill(Array(9).fill({ value: "empty" }));
```

fzaninotto (changes requested):

src/gameState.spec.ts
atilbian: We already have seen this before(https://github.com/marmelab/Hex/pull/4#discussion_r802722179) but has answered it's doesn't work as is so we want to wait for https://trello.com/c/i8YUKm6Q/5-05-as-norbert-i-want-to-be-able-to-play-on-any-size-of-board to check for a valid solution. Is it ok for you ?

src/game.ts
atilbian: It's coming soon!

src/gameState.ts
```diff
@@ -7,25 +7,34 @@ export interface GameState {
...
 export interface Cell {
-  value: "empty" | "black";
+  value: "empty" | "black" | "white";
 }
 
-export function playerHasWon(gameState: GameState): boolean {
-  const hexBoardGraph = createGraphFromGameState(gameState);
-  return doesPathExistForPlayer(hexBoardGraph);
-}
+export type stoneColor = "black" | "white";
```
slax57: type name should start with a capital letter

src/gameState.ts
```diff
@@ -7,25 +7,34 @@ export interface GameState {
...
-  return !!(pathBlack || pathWhite);
+function doesPathExistForPlayer(
+  hexBoardGraph: HexBoardGraph,
+  stoneColor: stoneColor
+): boolean {
+  if (stoneColor == "black") {
+    return !!doesPathExist(hexBoardGraph, "black-start", "black-end");
+  } else {
+    return !!doesPathExist(hexBoardGraph, "white-start", "white-end");
+  }
```
slax57: Actually I made a mistake here, the `doesPathExist()` method should return a boolean, hence, the `!!` syntax should be in this method rather than in the method `doesPathExistForPlayer()`. Can you move it please?

slax57 (changes requested):

slax57 (approved):

PR #10, opened 2022-02-10
Feature 9 implement mouse input in terminal

slax57:
- [x] Add hover on clickable objects
- [x] See if I can add some tests to cover this part as well (other tests have already been updated)

Trello card : https://trello.com/c/8T3rPYTn/9-2-as-norbert-i-find-that-placing-a-stone-from-its-coordinates-is-pretty-cumbersome


src/gui.ts
```diff
@@ -0,0 +1,241 @@
...
+    smartCSR: true,
+  });
+  // Quit on Escape, q, or Control-C.
+  screen.key(["escape", "q", "C-c"], function (ch, key) {
+    return process.exit(0);
+  });
+  return screen;
+}
+
+export function renderBoardAndLoop(
```
slax57: Need to split this method into submethods (it's too long)

src/gameState.ts
```diff
@@ -12,6 +13,22 @@ export interface Cell {
 
 export type StoneColor = "black" | "white";
 
+export function whoHasWon(gameState: GameState): StoneColor {
```
fzaninotto: I tend to prefer names that help to form a sentence when used in the code. 

e.g. `if(getWinner(gameState) === BLACK)` is more natural to read than `if(whoHasWon(gameState) === BLACK)`

src/gameState.ts
```diff
@@ -12,6 +13,22 @@ export interface Cell {
...
+    return stoneColor;
+  }
+  stoneColor = "white";
+  if (playerHasWon(gameState, stoneColor)) {
+    return stoneColor;
+  }
+  return null;
+}
+
+export function someoneWon(gameState: GameState): boolean {
```
fzaninotto: rename to `gameIsFinished` or `gameIsWon` ?

src/gameState.ts
```diff
@@ -55,15 +72,26 @@ export function doesCellExist(
   return true;
 }
 
+/**
+ * Returns true if a given cell has a stone placed on it.
+ * If a `stoneColor` is defined, this function will return true only if the cell
+ * has a stone on it and this stone is of the given color.
+ */
 export function doesCellHaveStone(
```
fzaninotto: I'd rename this on, too, to `cellHasStone`, it reads better in 

```js
if (cellHasStone(cell)) {
```

src/gui.ts
```diff
@@ -0,0 +1,245 @@
...
+    boardLayout.append(createWinMsg(gameState));
+    boardLayout.append(createHelpMsg(gameState));
+  }
+
+  screen.append(boardLayout);
+
+  screen.render();
+}
+
+function createBoxForPlayerTurn(
```
fzaninotto: createCurrentPlayerMsg? would be more consistent with the others

src/gui.ts
```diff
@@ -0,0 +1,245 @@
...
+    left: TEXT_LEFT_OFFSET,
+    content: "Press q to exit",
+    tags: true,
+    style: {
+      fg: "gray",
+    },
+  });
+}
+
+function createBoxForCell(
```
fzaninotto: createCell?

src/render.spec.ts
```diff
@@ -1,103 +0,0 @@
-import expect from "expect";
```
fzaninotto, line 1: Sad to see 1 day of work being trashed... But that's what we do! This may help you to understand how much time to spend on a feature.

fzaninotto (changes requested):

src/gui.ts
slax57: Done! 🙂

src/gameState.ts
```diff
@@ -57,12 +74,21 @@ export function doesCellExist(
   return true;
 }
 
-export function doesCellHaveStone(
+/**
```
atilbian, line 78: Could be splitted into 2 funcs : cellHasStoneOfColor and cellHasAnyStone to be more explicit without the need of comments.

src/gameState.ts
```diff
@@ -74,24 +100,24 @@ export function generateNewBoard(size: number): GameState {
         .map(() => deepCloneObject(DEFAULT_CELL_VALUE))
     );
 
-  return { board: newBoard };
+  return {
+    turn: "white",
+    board: newBoard
+  };
```
atilbian: Seems to me that it's not the goal of generateNewBoard to defined the starting stone. It could be defined at the calling level

src/parseConfigFile.ts
```diff
@@ -28,7 +28,10 @@ export function parseConfigFile(
   for (const row of data) {
     board.push(parseRow(row, rowsCount, filePath));
   }
-  const gameState = { board: board };
```
atilbian, line 31: It's not always the turn to white stone, it depends at least of the number of stones already on board.

src/gui.ts
```diff
@@ -0,0 +1,280 @@
+import * as blessed from "blessed";
```
atilbian, line 1: This file is much bigger than the others! But I don't know if it's time to split it with the little time left.

src/gui.ts
```diff
@@ -0,0 +1,280 @@
...
+    smartCSR: true,
+  });
+  // Quit on Escape, q, or Control-C.
+  screen.key(["escape", "q", "C-c"], function (ch, key) {
+    return process.exit(0);
+  });
+  return screen;
+}
+
+export function renderBoardAndLoop(
```
atilbian: Don't know if the loop part in name is still relevant

src/gui.ts
```diff
@@ -0,0 +1,280 @@
...
+    createColumnHeaderLabel(boardLayout, x);
+    // Create top edge
+    createBoardTopEdge(boardLayout, x);
+    // Create bottom edge
+    createBoardBottomEdge(boardLayout, gameState, x);
+  });
+
+  // For each row
+  gameState.board.forEach((row, y) => {
+    // Create row header label
```
atilbian: For me this kind of comments is redundant with the code

src/gui.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  screen.append(boardLayout);
+  screen.render();
+}
+
+function createGameTitleLabel() {
+  return blessed.text({
+    top: 0,
+    left: TEXT_LEFT_OFFSET,
+    content: "The Game of Hex",
```
atilbian, line 73: Nothing to change here but just a thought, we have more and more texts and I think that one day it will be useful to extract all these texts in separate files to allow to maintain them more easily

src/gui.ts
```diff
@@ -0,0 +1,280 @@
...
+  });
+}
+
+function createCurrentPlayerMsg(
+  gameState: GameState
+): blessed.Widgets.TextElement {
+  return blessed.text({
+    top: "center",
+    left: TEXT_LEFT_OFFSET,
+    content: `{bold}${gameState.turn}{/bold} turn to play`,
```
atilbian: Could be in uppercase to be more easy to read ? Or in another color like black/white(but need to change background if i remember well)

fzaninotto:
one question: how can you test the new renderer?

src/parseConfigFile.ts
slax57: Done! 🙂

slax57:
> one question: how can you test the new renderer?

Indeed, all code implying the blessed lib is currently uncovered. I believe that the risk it represents is mitigated by the fact that all the game logic behaviors are still covered by tests, but it would be nice to ensure that no regression happens to our beautiful GUI... I believe we would probably need another lib to test the blessed features, since blessed itself does not provide any way to do automated tests on the resulting GUI.

I suggest we discuss this together and agree on whether this needs to be done or not, an in any case, I believe this PR can be merged since it's already big and has been tested manually.

PR #13, opened 2022-02-14
Feature serve simple board state in html

atilbian:
- [x] Extract cli and common files into specific folders
- [x] Init a nest app
- [x] Create a controller displaying a board state(from config file on srv ~or a query parameter~)
- [x] Test the controller
- [x] ~Update ci script to make it test nest app too~ nothing to change here
- [x] Update the readme to explain how to start the web project

Related to : https://trello.com/c/MYewUZqP/38-as-norbert-i-want-to-be-able-to-see-a-particular-state-of-a-hex-game

## Demo
demo-feature-simple-board-state-in-html


test/app.e2e-spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    app = moduleFixture.createNestApplication();
+    await app.init();
+  });
+
+  it('/ (GET)', () => {
+    return request(app.getHttpServer())
+      .get('/')
+      .expect(200)
+      .expect('Hello World!');
+  });
```
slax57: Given your new `src/app.controller.ts` file, I would be surprised if this test still worked. Can you confirm it doesn't? And remove it or replace it with a working one if applicable.

.gitignore
```diff
@@ -1,4 +1,33 @@
...
+logs
+*.log
+npm-debug.log*
+pnpm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+lerna-debug.log*
+
+# OS
+.DS_Store
```
fzaninotto: os-specific, this one should be in the developer .gitignore, not in the project

.gitignore
```diff
@@ -1,4 +1,33 @@
...
+
+# Tests
+/coverage
+/.nyc_output
+
+# IDEs and editors
+/.idea
```
fzaninotto: same

gameStateFile.json
```diff
@@ -0,0 +1,35 @@
+[
+  [
+    {
+      "value": 1
```
fzaninotto, line 4: I don't usually discuss formatting, but the limit to 10 columns makes me wonder why you chose everyone to scroll a lot while we have large screens ? ;)

src/app.controller.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  describe('root', () => {
+    it('should return the current game state from the file"', () => {
+      expect(appController.getBoardStateFromFile().gameState).toEqual(
+        parseGameStateFromMultilineString(`
+        ⬢ ⬡ ⬢
+         ⬡ ⬡ ⬡
+          ⬡ ⬡ ⬡
+                `),
```
fzaninotto: I wonder why you need these extra blanks in the last line? It looks weird.

fzaninotto (commented):
Awesome! Just a few comments here and there.

test/app.e2e-spec.ts
atilbian: Good catch ! I will just remove the default test for now because no e2e tests are expected for now.

.gitignore
atilbian: Ok, i will remove all of these too :
```# IDEs and editors
/.idea
.project
.classpath
.c9/
*.launch
.settings/
*.sublime-workspace

# IDE - VSCode
.vscode/*
```

gameStateFile.json
atilbian: Seems like it's the default behaviour with prettier ! Do we need to change it ?

fzaninotto: no, if prettier does it, let's keep it.

PR #15, opened 2022-02-15
Make app deployable with aws elastic beanstalk

atilbian:
- [x] Add required npm scripts to help deploying app
- [x] Add and configure procfile
- [x] Deploy app manually : http://awseb-awseb-yru6x8bojx9q-300182659.us-east-1.elb.amazonaws.com/
- [x] Use github action/travis CI to deploy app automatically
- [x] Generate an AWS token not linked to a personal account
- [x] Configure AWS_ACCESS_KEY in github env var **(encrypted)**

## Links
https://github.com/marketplace/actions/beanstalk-deploy#:~:text=Beanstalk%20Deploy%20is%20a%20GitHub,that%20version%20to%20the%20environment
https://towardsdev.com/how-to-deploy-a-nestjs-app-to-aws-elastic-beanstalk-with-added-https-and-ci-cd-2d35c319e9f4
https://github.com/marmelab/casquette/blob/master/.travis.yml#L37

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,9 @@
   "author": "",
   "private": true,
   "license": "UNLICENSED",
+  "engines": {
+    "node": "^12"
+  },
   "scripts": {
     "prebuild": "rimraf dist",
     "build": "nest build",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
     "test:e2e": "jest --config ./test/jest-e2e.json",
     "start:cli": "ts-node src/cli/index.ts",
-    "dev:cli": "nodemon --quiet --watch 'src/**/*.ts' --exec 'ts-node' src/cli/index.ts"
+    "dev:cli": "nodemon --quiet --watch 'src/**/*.ts' --exec 'ts-node' src/cli/index.ts",
+    "deploy": "npm ci && npm run build && npm run start:prod"
   },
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -20,7 +23,8 @@
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
     "test:e2e": "jest --config ./test/jest-e2e.json",
     "start:cli": "ts-node src/cli/index.ts",
-    "dev:cli": "nodemon --quiet --watch 'src/**/*.ts' --exec 'ts-node' src/cli/index.ts"
+    "dev:cli": "nodemon --quiet --watch 'src/**/*.ts' --exec 'ts-node' src/cli/index.ts",
+    "deploy": "npm ci && npm run build && npm run start:prod"
```
slax57, line 27: Just a small question, is there a good reason to use npm here and not yarn? I read it's a good practice not to mix up package managers 🙂

atilbian: Yarn isn't installed by default on the aws EB/node.js so i choose the faster solution ! But seems the good way to go to use only yarn. We could create an issue related to this and merge this branch now ?

slax57: I don't know exactly what it would take to install yarn on the aws EB/node.js, so I can't say if it's worth bothering.
I believe the most important thing is not to mixup yarn and npm while editing projet dependencies, but for run and deploy commands I don't think it's that harmful to use npm.

TLDR: Let's forget about this and merge as is. I wouldn't event create an issue for this in the end.

PR #16, opened 2022-02-15
Add graphical hexagonal board representation using HTML and CSS

slax57:
## Checklist

- [x] Add graphical hexagonal board representation using HTML and CSS
- [x] Add black and white edges
- [x] Replace hacky mapping of the value `"empty"` to `"#eea"` in Controller by a HBS Helper
- [x] Add E2E tests

## Demo

hex-board-html-css

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -1,0 +1,83 @@
+/* Great source: https://css-tricks.com/hexagons-and-beyond-flexible-responsive-grid-patterns-sans-media-queries/ */
+
+#board {
+    display: grid;
+    place-self: center;
+    justify-content: center;
+    margin: 20px;
+    padding: 20px;
+    background: #333;
+}
+
+.main {
+    display: flex;
+    flex-direction: column;
+    /* size  */
+    --s: 30px;
+    /* margin */
+    --m: 3px;
+    /* height equivalent to 2 overlapping hexagons, used to move every second row */
+    --f: calc(1.732 * var(--s) + 4 * var(--m) - 1px);
+}
+
+.container {
+    /*disable white space between inline block element */
+    font-size: 0;
+    width: calc((var(--nr) + (var(--nr)) / 2)*(var(--s) + 2*var(--m)));
+    padding-bottom: calc(var(--s) * 0.2885 * 2);
+    padding-top: calc(var(--s) * 0.2885 * 2);
+}
+
+.container div {
+    width: var(--s);
+    margin: var(--m);
+    height: calc(var(--s)*1.1547);
+    display: inline-block;
+    font-size: initial;
+    clip-path: polygon(0% 25%, 0% 75%, 50% 100%, 100% 75%, 100% 25%, 50% 0%);
+    margin-bottom: calc(var(--m) - var(--s)*0.2885);
+}
+
+.container::before,
+.container i {
+    content: "";
+    width: calc((var(--s) + 2*var(--m))*((var(--nr))/2));
+    float: left;
+    height: calc((var(--f)*(var(--nr))/2) + var(--s)*0.2885);
+    shape-outside: linear-gradient(to top right, #000 50%, #0000 0);
+}
+
+.container i {
+    float: right;
+    shape-outside: linear-gradient(to bottom left, #000 calc(50% - 1*var(--s)), #0000 0);
+}
+
+#top-edge, #bottom-edge, #left-edge, #right-edge {
+    position: absolute;
+    width: calc((var(--s) + 2*var(--m))*(var(--nr)));
+    height: calc(var(--s)*0.2885*2);
+    background: white;
+    clip-path: polygon(2% 0%, 98% 0%, 100% 50%, 98% 100%, 2% 100%, 0% 50%);
+}
+
+#top-edge {
+    transform: translate(calc(var(--s) / 2), calc(0px - 2*var(--m)));
+}
+
+#bottom-edge {
+    transform: translate(calc((var(--s) + 2*var(--m))*((var(--nr) - 1)/2) + var(--s) / 1.5), calc(var(--f)*var(--nr)/2 + var(--s)/1.5 + var(--s) * 0.2885 * 2)) rotate(180deg);
+}
+
+#left-edge, #right-edge {
+    background: black;
+    transform-origin: left;
+}
+
+#left-edge {
+    transform: scaleY(-1) rotate(300deg) translate(calc(0px - 2*var(--m) + var(--s) / 2), calc(0px - var(--s)*0.2885*2.5));
+}
+
+#right-edge {
+    transform-origin: right;
+    transform: scaleY(-1) rotate(120deg) translate(calc(0px - 2*var(--m) - var(--s) / 1.5), calc(0px + var(--s)*0.2885*2 - var(--s) / 0.9));
+}
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -11,22 +11,23 @@
 
 .hex-board-frame {
     display: flex;
-    flex-direction: column;
     /* width of one hexagon  */
     --hexwidth: 30px;
     /* margin around hexagons */
     --hexmargin: 3px;
     /* condensed height equivalent to the height of 2 overlapping hexagons (used to move every second row) */
     --hexcondheight: calc(1.732 * var(--hexwidth) + 4 * var(--hexmargin) - 1px);
+    /* additional height of one hexagon compared to its width (to have nice round hexagons)  */
+    --hexaddheight: calc(var(--hexwidth) * 0.2885);
     /* last variable `--rows` holds the number of rows and is injected by the template */
 }
 
 .hex-board-container {
     /*disable white space between inline block element */
     font-size: 0;
-    width: calc((var(--rows) + (var(--rows)) / 2)*(var(--hexwidth) + 2*var(--hexmargin)));
-    padding-bottom: calc(var(--hexwidth) * 0.2885 * 2);
-    padding-top: calc(var(--hexwidth) * 0.2885 * 2);
+    width: calc((var(--rows) * 1.5)*(var(--hexwidth) + 2*var(--hexmargin)));
+    padding-bottom: calc(var(--hexaddheight) * 2);
+    padding-top: calc(var(--hexaddheight) * 2);
 }
 
 .hex-board-container div {
@@ -34,9 +35,8 @@
     margin: var(--hexmargin);
     height: calc(var(--hexwidth)*1.1547);
     display: inline-block;
-    font-size: initial;
     clip-path: polygon(0% 25%, 0% 75%, 50% 100%, 100% 75%, 100% 25%, 50% 0%);
-    margin-bottom: calc(var(--hexmargin) - var(--hexwidth)*0.2885);
+    margin-bottom: calc(var(--hexmargin) - var(--hexaddheight));
 }
 
 .hex-board-container::before,
@@ -44,7 +44,7 @@
     content: "";
     width: calc((var(--hexwidth) + 2*var(--hexmargin))*((var(--rows))/2));
     float: left;
-    height: calc((var(--hexcondheight)*(var(--rows))/2) + var(--hexwidth)*0.2885);
+    height: calc((var(--hexcondheight)*(var(--rows))/2) + var(--hexaddheight));
     shape-outside: linear-gradient(to top right, #000 50%, #0000 0);
 }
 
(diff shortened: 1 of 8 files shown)


fzaninotto:
Would you mind rebasing the PR so that the diff becomes easier to review?

public/css/style.css
```diff
@@ -0,0 +1,83 @@
+/* Great source: https://css-tricks.com/hexagons-and-beyond-flexible-responsive-grid-patterns-sans-media-queries/ */
+
+#board {
```
fzaninotto: in general, we don't use ids very much in CSS, mostly classes. Because there may be more than one board per page...

public/css/style.css
```diff
@@ -0,0 +1,83 @@
...
+    display: grid;
```
fzaninotto, line 4: 👍 that's the way to go!

public/css/style.css
```diff
@@ -0,0 +1,83 @@
...
+    place-self: center;
+    justify-content: center;
+    margin: 20px;
+    padding: 20px;
+    background: #333;
+}
+
+.main {
```
fzaninotto: this class name isn't very explanatory. Try to find a better name for it.

public/css/style.css
```diff
@@ -0,0 +1,83 @@
...
+    margin: var(--m);
+    height: calc(var(--s)*1.1547);
+    display: inline-block;
+    font-size: initial;
+    clip-path: polygon(0% 25%, 0% 75%, 50% 100%, 100% 75%, 100% 25%, 50% 0%);
+    margin-bottom: calc(var(--m) - var(--s)*0.2885);
+}
+
+.container::before,
+.container i {
```
fzaninotto: I'm not sure what this represents, I'm a bit afraid that this will make hover on hexagons fail.

public/css/style.css
```diff
@@ -0,0 +1,83 @@
...
+    height: calc((var(--f)*(var(--nr))/2) + var(--s)*0.2885);
+    shape-outside: linear-gradient(to top right, #000 50%, #0000 0);
+}
+
+.container i {
+    float: right;
+    shape-outside: linear-gradient(to bottom left, #000 calc(50% - 1*var(--s)), #0000 0);
+}
+
+#top-edge, #bottom-edge, #left-edge, #right-edge {
```
fzaninotto: same, please use classes

public/css/style.css
```diff
@@ -0,0 +1,83 @@
...
+
+#top-edge {
+    transform: translate(calc(var(--s) / 2), calc(0px - 2*var(--m)));
+}
+
+#bottom-edge {
+    transform: translate(calc((var(--s) + 2*var(--m))*((var(--nr) - 1)/2) + var(--s) / 1.5), calc(var(--f)*var(--nr)/2 + var(--s)/1.5 + var(--s) * 0.2885 * 2)) rotate(180deg);
+}
+
+#left-edge, #right-edge {
```
fzaninotto: even if that means more code, I'd prefer that you put these props into the respective left and right edge styles. Eases review and maintenance.

public/css/style.css
```diff
@@ -0,0 +1,83 @@
...
+    margin: 20px;
+    padding: 20px;
+    background: #333;
+}
+
+.main {
+    display: flex;
+    flex-direction: column;
+    /* size  */
+    --s: 30px;
```
fzaninotto: You should use a more explicit variable name

public/css/style.css
```diff
@@ -0,0 +1,83 @@
...
+    /* margin */
+    --m: 3px;
+    /* height equivalent to 2 overlapping hexagons, used to move every second row */
+    --f: calc(1.732 * var(--s) + 4 * var(--m) - 1px);
+}
+
+.container {
+    /*disable white space between inline block element */
+    font-size: 0;
+    width: calc((var(--nr) + (var(--nr)) / 2)*(var(--s) + 2*var(--m)));
```
fzaninotto: what's this --nr var?

Edit: I just saw it in the template. You should probbaly add a CSS comment in this file explaining what --nr is, what type of value it can contain... and find a better name ;)

src/main.ts
```diff
@@ -2,13 +2,24 @@ import { NestFactory } from '@nestjs/core';
 import { NestExpressApplication } from '@nestjs/platform-express';
 import { AppModule } from './app.module';
 import { join } from 'path';
+import * as hbs from 'hbs';
```
fzaninotto: I prefer an import of the actual function, but that's debatable. 

```jsx
import { handlebars } from 'hbs';
```

I guess it's mostly a matter of taste. 

But since some imports are named and some imports are wildcards, we will always ask ourselfs which type of import we should do in this progral. 

My suggestion: choose one style and stick to it.

src/main.ts
```diff
@@ -2,13 +2,24 @@ import { NestFactory } from '@nestjs/core';
...
   const app = await NestFactory.create<NestExpressApplication>(AppModule);
   app.useStaticAssets(join(__dirname, '..', 'public'));
   app.setBaseViewsDir(join(__dirname, '..', 'views'));
+  registerHandlebarsHelpers();
   app.setViewEngine('hbs');
   await app.listen(3000);
 }
 
+function registerHandlebarsHelpers() {
+  hbs.handlebars.registerHelper('ifCond', function (v1, v2, options) {
```
fzaninotto: The #if helper is built-in, why do you need another one?

https://handlebarsjs.com/guide/builtin-helpers.html#if

views/index.hbs
```diff
@@ -4,17 +4,26 @@
...
-    {{/each}}
+    <div class="main" style="--nr: {{gameState.board.length}}">
+      <div id="left-edge"></div>
+      <div id="right-edge"></div>
+      <div id="top-edge"></div>
+      <div id="bottom-edge"></div>
+      <div class="container">
+        <i></i>
+        {{#each gameState.board}}
+        {{#each this}}
```
fzaninotto: please add indentation for the second each block

fzaninotto (changes requested):

public/css/style.css
slax57: Actually its a suggestion from https://css-tricks.com/hexagons-and-beyond-flexible-responsive-grid-patterns-sans-media-queries/

They use a `<i>` element because they can't have two `::before` elements, but we need two of them to achieve the desired effect.

Theses elements are used mainly to add a `shape-outside` property, allowing to offset the hex cells progressively.

Here is a preview of what it looks like, using Chrome DevTools:

2022-02-15 15_26_13-Hex

I still haven't tried to add a hover effect, but I see no reason why this would make it fail so far.

src/main.ts
slax57: Unless I'm missing something, HandleBars can only reference the view model attributes, but cannot execute any kind of JS code 🙁 .

If `value` already contains a boolean (or something you expect to be truthy or falsy), then you may use hbs's built-in `#if` helper, but if you intend to do something like `{{#if value=='empty'}}`, you need to create your own helper for this 🙁

src/main.ts
```diff
@@ -2,13 +2,24 @@ import { NestFactory } from '@nestjs/core';
...
 }
 
+export function registerHandlebarsHelpers() {
+  handlebars.registerHelper('ifCond', function (v1, v2, options) {
+    if (v1 === v2) {
+      return options.fn(this);
+    }
+    return options.inverse(this);
+  });
+}
```
atilbian: It could be extracted into a handlebars helper module because i think it will soon be used into your e2e tests too !

src/main.ts
```diff
@@ -2,13 +2,24 @@ import { NestFactory } from '@nestjs/core';
...
   const app = await NestFactory.create<NestExpressApplication>(AppModule);
   app.useStaticAssets(join(__dirname, '..', 'public'));
   app.setBaseViewsDir(join(__dirname, '..', 'views'));
+  registerHandlebarsHelpers();
   app.setViewEngine('hbs');
   await app.listen(3000);
 }
 
+export function registerHandlebarsHelpers() {
+  handlebars.registerHelper('ifCond', function (v1, v2, options) {
```
atilbian: value1 and value2 to be more explicit ?

views/index.hbs
```diff
@@ -4,17 +4,26 @@
...
+    <div class="hex-board-frame" style="--rows: {{gameState.board.length}}">
+      <div class="left-edge"></div>
+      <div class="right-edge"></div>
+      <div class="top-edge"></div>
+      <div class="bottom-edge"></div>
+      <div class="hex-board-container">
+        <i></i>
+        {{#each gameState.board}}
+          {{#each this}}
+            <div style="background: {{#ifCond this.value 'empty'}}#eea{{else}}{{this.value}}{{/ifCond}}">
```
atilbian, line 21: The helper method could map all our Cell enum values to selected css colors ? So we could change it easily or update the enum without breaking the colors ?

public/css/style.css
```diff
@@ -0,0 +1,82 @@
...
+
+.hex-board-frame {
+    display: flex;
+    flex-direction: column;
+    /* width of one hexagon  */
+    --hexwidth: 30px;
+    /* margin around hexagons */
+    --hexmargin: 3px;
+    /* condensed height equivalent to the height of 2 overlapping hexagons (used to move every second row) */
+    --hexcondheight: calc(1.732 * var(--hexwidth) + 4 * var(--hexmargin) - 1px);
```
atilbian, line 19: Maaaaaaaagic numbers ! (1.732,4,0.2885, 2, 1.1547)

public/css/style.css
```diff
@@ -0,0 +1,82 @@
...
+.hex-board-container i {
+    float: right;
+    shape-outside: linear-gradient(to bottom left, #000 calc(50% - 1*var(--hexwidth)), #0000 0);
+}
+
+.top-edge, .bottom-edge, .left-edge, .right-edge {
+    position: absolute;
+    width: calc((var(--hexwidth) + 2*var(--hexmargin))*(var(--rows)));
+    height: calc(var(--hexwidth)*0.2885*2);
+    background: white;
```
atilbian: Could be moved into .top-edge and .bottom-edge classes. But i think it could be improved by creating a new "white" and "black" class which only define the corresponding background color, then use these classes in the html.

public/css/style.css
```diff
@@ -0,0 +1,82 @@
...
+    place-self: center;
+    justify-content: center;
+    margin: 20px;
+    padding: 20px;
+    background: #333;
+}
+
+.hex-board-frame {
+    display: flex;
+    flex-direction: column;
```
atilbian: I don't think it's used anywhere because children are absolutely positioned here !

public/css/style.css
```diff
@@ -0,0 +1,82 @@
...
+    background: white;
+    clip-path: polygon(2% 0%, 98% 0%, 100% 50%, 98% 100%, 2% 100%, 0% 50%);
+}
+
+.top-edge {
+    transform: translate(calc(var(--hexwidth) / 2), calc(0px - 2*var(--hexmargin)));
+}
+
+.bottom-edge {
+    transform: translate(calc((var(--hexwidth) + 2*var(--hexmargin))*((var(--rows) - 1)/2) + var(--hexwidth) / 1.5), calc(var(--hexcondheight)*var(--rows)/2 + var(--hexwidth)/1.5 + var(--hexwidth) * 0.2885 * 2)) rotate(180deg);
```
atilbian: I think rotate(180deg) is useless here

public/css/style.css
```diff
@@ -0,0 +1,82 @@
...
+    padding-bottom: calc(var(--hexwidth) * 0.2885 * 2);
+    padding-top: calc(var(--hexwidth) * 0.2885 * 2);
+}
+
+.hex-board-container div {
+    width: var(--hexwidth);
+    margin: var(--hexmargin);
+    height: calc(var(--hexwidth)*1.1547);
+    display: inline-block;
+    font-size: initial;
```
atilbian: I think it's useless too

public/css/style.css
```diff
@@ -0,0 +1,82 @@
...
+    --hexmargin: 3px;
+    /* condensed height equivalent to the height of 2 overlapping hexagons (used to move every second row) */
+    --hexcondheight: calc(1.732 * var(--hexwidth) + 4 * var(--hexmargin) - 1px);
+    /* last variable `--rows` holds the number of rows and is injected by the template */
+}
+
+.hex-board-container {
+    /*disable white space between inline block element */
+    font-size: 0;
+    width: calc((var(--rows) + (var(--rows)) / 2)*(var(--hexwidth) + 2*var(--hexmargin)));
```
atilbian: var(--rows) + (var(--rows)) / 2 could be simplified to var(--rows)*1.5

public/css/style.css
slax57: Magic numbers are explained in the referenced website.

I've created a new variable to store the 0.2885 that appeared multiple times though.

public/css/style.css
slax57: indeed, it's a leftover from a previous attempt, thanks for pointing this out

public/css/style.css
slax57: same here, leftover code from previous shape (when it was not symmetrical). will remove it

views/index.hbs
slax57: Since it's a _nice to have_ improvement, and I've already spent way too much time on this feature, I've opened an Issue to keep track of your suggestion (which I approve). And I suggest we merge this PR as is.

Issue: https://github.com/marmelab/Hex/issues/19

fzaninotto (approved):

PR #17, opened 2022-02-15
Allow player to add stones on board

atilbian:
- [x] Add e2e tests
- [x] Add uts
- [x] Add hover effect on empty cells
- [x] Make template generate each cell as a form with a querystring representing the current board state + coords of current cell
- [x] Make the controller extract board state + coords of clicked cell from querystring

Related to : https://trello.com/c/HoL63RFd/1-05-as-norbert-i-want-to-be-able-to-add-a-stone-anywhere-on-the-board

## Demo
demo-feature-add-stones


src/app.controller.ts
```diff
@@ -1,14 +1,43 @@
...
+  getBoard(
+    @Query('gameState') playerGameState?: string,
+    @Query('x') x?: number,
+    @Query('y') y?: number,
+  ): { gameState: GameState } {
+    let gameState = playerGameState
+      ? (parse(playerGameState) as unknown as GameState)
+      : this.appService.initNewGameState();
+    if (x && y) {
+      gameState = this.appService.updateGameState(gameState, {
```
fzaninotto: there is a misconception about mutations here, let's speak about it when you have a moment.

src/app.service.ts
```diff
@@ -1,16 +1,34 @@
...
     const gameState = parseGameStateFromFile(
       join(__dirname, configPathFromDistDir),
     );
-    return { gameState };
+    return gameState;
+  }
+
+  updateGameState(gameState: GameState, coordinates: Coordinates): GameState {
+    try {
+      updateGameState(gameState, coordinates);
```
fzaninotto: `updateGameState` is supposed to return a new game state, not to modify its parameters (which is considered a bad practice in functional programming, and is even forbiddent by some runtimes)

views/index.hbs
```diff
@@ -8,6 +8,17 @@
 </head>
 
 <body>
+  {{#each gameState.board}}
+  {{#each this}}
+  <form method="get" action="/">
```
fzaninotto: I don't understand why you "draw" the board twice, once with a hidden form and once with divs. Why not draw it once with a visible form instead?

fzaninotto (changes requested):

src/app.service.ts
atilbian: Oh yes, i forgot to use the game state returned by updateGameState.

Is something like this okay for you ?
```
try {
  const updatedGameState = updateGameState(gameState, coordinates);
  return updatedGameState;
} catch {
  return gameState;
}
```

views/index.hbs
atilbian: This is a solution I tried to make something work quickly. My first attempts in the way you suggest broke the style of the board. But i can retry now to merge cells+forms.

src/app.controller.ts
atilbian: Ok !

src/app.service.ts
```diff
@@ -1,16 +1,35 @@
...
-    return { gameState };
+    return gameState;
+  }
+
+  updateGameState(gameState: GameState, coordinates: Coordinates): GameState {
+    try {
+      const updatedGameState = updateGameState(gameState, coordinates);
+      return updatedGameState;
+    } catch {
+      return gameState;
```
arimet: Isn't it weird to return the old gamestate if you had a problem? Shouldn't we return an error?

public/css/style.css
```diff
@@ -1,5 +1,9 @@
 /* Great source: https://css-tricks.com/hexagons-and-beyond-flexible-responsive-grid-patterns-sans-media-queries/ */
 
+form {
```
arimet: Since your css is global for the moment, I think you have to be more precise in the selector

slax57 (approved):

PR #20, opened 2022-02-16
Feature persist games in a Postgres database

slax57:
## Feature

Ability to persist games in a Postgres database, in order to resume it later.

Trello card: https://trello.com/c/UJhASbow/10-1-as-norbert-i-want-to-continue-a-game-i-started-in-the-past

## Checklist

- [x] Use **TypeORM** as ORM in NestJS
- [x] Add postgres drivers
- [x] Create entities to persist games in db
- [x] Create a Dockerfile for the NestJS app
- [x] Create a Docker Compose file to run the NestJS server and postgreSQL server as services, as well as a pgadmin instance (for easier debugging)
- [x] Update Readme file with commands and environment variables
- [ ] ~Create Dockerfile for the CLI app~
- [x] Use `gamesRepository` in a simple proof of concept (insert and select)
- [x] Create a _start new game_ page and operation
- [x] Store each game created this way in database, and persist game state after each move
- [x] Implement ability to resume a game in the future
- [ ] ~Update e2e tests~
- [ ] ~Fix Cypress tests~
- [x] Update AWS deployment pipeline


src/app.service.ts
```diff
@@ -2,15 +2,43 @@ import { Injectable } from '@nestjs/common';
...
+
+  findGameById(id: number): Promise<Game> {
+    return this.gamesRepository.findOne(id);
+  }
+
+  async createNewGame(size: number): Promise<number> {
+    if (!size) {
+      size = 5;
+    } else {
+      size = +size;
```
arimet, line 50: useless since you can only pass a number to your function

src/app.service.ts
```diff
@@ -2,15 +2,43 @@ import { Injectable } from '@nestjs/common';
...
     return { gameState };
   }
+
+  findGameById(id: number): Promise<Game> {
+    return this.gamesRepository.findOne(id);
+  }
+
+  async createNewGame(size: number): Promise<number> {
+    if (!size) {
+      size = 5;
```
arimet: Magical number, use DEFAULT_SIZE maybe ?

src/app.service.ts
```diff
@@ -2,15 +2,43 @@ import { Injectable } from '@nestjs/common';
...
+    } else {
+      size = +size;
+    }
+    const game = await this.gamesRepository.save(this.gamesRepository.create({
```
arimet: a lot of nested action, it makes the code harder to read

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+RUN mkdir -p /usr/app
+WORKDIR /usr/app
+
+COPY package.json yarn.lock tsconfig.json tsconfig.build.json start.sh ./
+COPY src ./src
+COPY public ./public
+COPY views ./views
+
+ADD https://github.com/ufoscout/docker-compose-wait/releases/download/2.2.1/wait /wait
```
arimet, line 13: why do you use docker-compose-wait?

src/app.service.ts
slax57: Apparently it is really needed, if I remove this line, then the behaviour of `Array(size)` changes

Dockerfile
slax57: It's a practice used to ensure the prerequisite services (here, the postgresql db) are ready to accept connections before we start our app

docker-compose.yaml
```diff
@@ -0,0 +1,66 @@
...
+      dockerfile: ./Dockerfile.nginx
+    ports:
+      - 80:80
+    depends_on:
+      - nestjs
+    restart: always
+
+volumes:
+  db-data:
+  pgadmin-data:
```
atilbian, line 66: Reminder : we could move all files related to the deployment part into a subfolder

ormconfig.json
```diff
@@ -0,0 +1,12 @@
...
+    "host": "127.0.0.1",
+    "port": 5432,
+    "username": "postgres",
+    "password": "postgres",
+    "database": "test",
+    "entities": [
+        "dist/**/*.entity.js"
+    ],
+    "synchronize": true
+}
```
atilbian, line 12: Reminder: find a way to avoid duplication (with the env var defined in docker). For example by creating a common file used with or without docker.

start.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+if [ "$NODE_ENV" == "production" ] ; then
+  yarn start:prod
+else
+  yarn start
```
atilbian, line 6: Reminder : find a way to develop into the nest docker container(with a shared volume containing the source code ?)

PR #22, opened 2022-02-17
Feature detect win

atilbian:
- [x] Add tests
- [x] Check for a winner at each turn and display it
- [x] Allow to go back to home page when game is finished
- [x] Wait for https://github.com/marmelab/Hex/pull/21 to be merged
Related to : https://trello.com/c/WiDF7hGp/47-05-as-norbert-given-someone-wins-i-want-the-game-to-show-it

## Demo
demo-detect-win


src/app.controller.ts
```diff
@@ -15,16 +15,12 @@ export class AppController {
   getBoardStateFromFile(
     @Query('x') x?: number,
     @Query('y') y?: number,
-  ): { gameState: GameState } {
+  ): { gameState: GameState; winner: StoneColor } {
```
slax57: why don't you add this attribute inside the GameState interface? imo it's part of the state of the game to know whether the game is won or not

src/app.controller.ts
```diff
@@ -15,16 +15,12 @@ export class AppController {
...
-      ? {
-          gameState: this.appService.updateGameState(gameState, {
-            x,
-            y,
-          }),
-        }
-      : { gameState };
+    const nextGameState =
+      x && y ? this.appService.updateGameState(gameState, { x, y }) : gameState;
+    const hasAWinner = getWinner(nextGameState);
```
slax57: In the same way, I feel like the win detection could be part of the `updateGameState()` function (inside a sub-function, of course)

src/app.controller.ts
atilbian: Ok good idea! I just choose the faster solution to implement. Because updating the game state interface will probably have an impact on the cli program too, so more stuff to work on. But let's try it !

atilbian: It's done(tell me if it's ok for you), and i also made some refactoring to use this new prop when possible.

slax57 (approved):

src/app.controller.ts
slax57: that's great, well done! 🙂

PR #33, opened 2022-02-18
Update targeted elastic beanstalk environnement

atilbian:
- [ ] Target our new Elastic Beanstalk environnement(old one is dead, long live the old one)!


slax57:
This PR is obsolete since we have: https://github.com/marmelab/Hex/pull/39

PR #34, opened 2022-02-21
Use docker to create a dev env

atilbian:
- [x] Allow to develop into docker
- [x] Hide env files from git
- [x] Update readme


README.md
```diff
@@ -32,12 +32,12 @@ Edit database connection params in `ormconfig.json` file.
 ### With Docker Compose
 
 ```bash
-$ docker-compose up -d
+$ docker-compose -f deployment/docker-compose.yaml up -d
```
slax57: Maybe mention that there are 2 docker-compose files for dev and prod

README.md
```diff
@@ -32,12 +32,12 @@ Edit database connection params in `ormconfig.json` file.
...
 ```
 
 #### Environment variables
 
-You may edit the following environment variables in `docker-compose.yaml`:
+You may edit the following environment variables in `.env.dev` or `.env.prod`:
```
slax57:
```suggestion
You may edit the following environment variables in `deployment/.env.dev` or `deployment/.env.prod`:
```

deployment/.env.dev.dist
```diff
@@ -0,0 +1,15 @@
...
+TYPEORM_PASSWORD=db_password
+TYPEORM_DATABASE=db_name
+TYPEORM_ENTITIES=dist/**/*.entity.js
+TYPEORM_SYNCHRONIZE="true"
+WAIT_HOSTS=postgres:5432
+NODE_ENV=development
+# db
+POSTGRES_USER=db_user
+POSTGRES_PASSWORD=db_password
+POSTGRES_DB=db_name
```
slax57: - I believe we could do stg more clever by using only 1 variable name for a given data (eg. POSTGRES_USER) and edit the docker-compose.yaml to do the _glue logic_ for us like so `TYPEORM_USERNAME: $POSTGRES_USER`
- Also, I think we should only put in this file variables that the user can edit safely without breaking the app. What I mean by that is that variables such as TYPEORM_ENTITIES, TYPEORM_SYNCHRONIZE or WAIT_HOSTS do not belong here according to me

deployment/Dockerfile.prod
```diff
@@ -0,0 +1,16 @@
+FROM node:12-alpine
+
+RUN mkdir -p /usr/app
+WORKDIR /usr/app
+
+COPY . .
```
slax57: I don't think we should copy everything in the prod image, cause we might end up copying files that are at best irrelevant and at worst that should not be accessible by the end user (like test files, or our own .env.prod file for example). I know that for now we deploy this image on AWS and we're supposed to be the only ones accessing it, but if one day we want to push this image to the Docker Hub for instance then currently we can't. I'd recommend sticking to listing the files and folders that are actually sources we want to copy.

deployment/docker-compose.prod.yaml
```diff
@@ -0,0 +1,54 @@
...
+    env_file:
+      - .env.prod
+    expose:
+      - 3000
+
+  postgres:
+    container_name: postgres
+    image: postgres:10.4
+    ports:
+      - "5432:5432"
```
slax57: Should map to TYPEORM_PORT, or TYPEORM_PORT should not be part of .env file

deployment/docker-compose.prod.yaml
```diff
@@ -0,0 +1,54 @@
...
+    image: dpage/pgadmin4:4.18
+    restart: always
+    env_file:
+      - .env.prod
+    ports:
+      - 8080:80
+    volumes:
+      - pgadmin-data-prod:/var/lib/pgadmin
+    depends_on:
+      - postgres
```
slax57: Didn't you remove pgadmin from the prod docker-compose? 🤔

deployment/docker-compose.yaml
```diff
@@ -0,0 +1,25 @@
...
+    ports:
+      - 3000:3000
+    volumes:
+      - ../:/usr/app/
+
+  postgres:
+    container_name: postgres
+    image: postgres:10.4
+    ports:
+      - 5432:5432
```
slax57: Same here concerning TYPEORM_PORT

deployment/.env.dev.dist
atilbian: Ok for the first point. And have you got an idea of how to define "unsafe variables" ? It could be set directly in the docker compose file !

deployment/Dockerfile.prod
atilbian: Yes i think it's safer to target manually each files/folders instead of copying everything.

deployment/docker-compose.prod.yaml
atilbian: I don't know if we can use an env value in the port definition, so we could remove these var and let the port definition in the docker compose file.

deployment/docker-compose.prod.yaml
atilbian: Nop, only from the dev version! Do you want to remove it from prod too? It doesn't seem to be useful for now, so we can clean it up.

deployment/.env.dev.dist
slax57: TYPEORM_ENTITIES and TYPEORM_SYNCHRONIZE could be added to the Dockerfile directly, while WAIT_HOSTS could be put inside the docker-compose.yaml, since it's about synchronization between multiple services

deployment/docker-compose.prod.yaml
slax57: I believe you can. But OK for now, hardcoding the port number in the docker-compose.yaml file seems fine to me.

deployment/docker-compose.prod.yaml
slax57: As discussed together, I'd rather have it in the dev docker-compose than the prod one 🙂

deployment/docker-compose.yaml
```diff
@@ -0,0 +1,25 @@
+version: "3"
```
floo51: Why is this file in the `deployment` folder when it's actually for dev?

deployment/docker-compose.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+  nestjs:
+    container_name: nestjs
+    build:
+      context: ../
+      dockerfile: ./deployment/Dockerfile
+    depends_on:
+      - postgres
+    env_file:
+      - .env.dev
```
floo51: Be very careful with sharing whole env variables with all containers.
There might be some secret env variable that we don't want to end-up in all containers, like front-end.

floo51 (changes requested):
The deployment folder seems like "the whole lot of stuff that makes the deployment work".
As for the rest of the code, we can organize it by domain (ie. domain = infra here) :

```
/backend
    package.json
    .env.*
    Dockerfile
/app
    package.json
    .env.* 
    Dockerfile
docker-compose.yml
docker-compose.prod.yml
Makefile
```

As for all the docker settings and conf files that are not directly linked to my project (e.g. nginx), I like to regroup them in their own `.docker` folder:

```
/.docker
    /nginx
         nginx.conf
    /postgres
         /data
         pg.conf
```

web-app/deployment/docker-compose.prod.yaml
```diff
@@ -0,0 +1,42 @@
+version: "3"
+services:
+
+  nestjs:
+    container_name: nestjs
```
floo51: Ideally you want to use a already built image in production instead of building from a Dockerfile.
The Docker image is called a deployment artifact, and can (ideally again) be used on any environment by customizing the env vars.

deployment/docker-compose.yaml
atilbian: So we need to create different env files for each service ?

deployment/docker-compose.yaml
atilbian: Good question. I will move it to the web-app root folder.

atilbian:
> 

What's behind "app" and "backend" in your example ? Seems 2 differents things for me so we don't need to share a common docker compose file between each ?

web-app/deployment/docker-compose.prod.yaml
atilbian: I'm not sure to see how i could mix this approach with the one suggested by Clement(with a multistage docker image). Could we have a talk about it ?

deployment/docker-compose.yaml
slax57: Sorry pal, merging all .env files into a single file was my idea 😦

web-app/deployment/docker-compose.yaml
```diff
@@ -0,0 +1,38 @@
+version: "3"
```
Clebiez: You can combine your docker-compose files in one line command in order to factorize the config

It's very useful to create some shared config in docker compose file.

example:

`docker-compose -f docker-compose.yml -f docker-compose.dev.yml` will merge this tow files to make only one config.
It means that you can have a main docker-compose defining all of your services and both docker-compose.dev.yml or docker-compose.prod.yml just have to override some params that you need for eachs envs (port, volumes etc) Don't hesitate to call me if you need a real world example, I have one :)

web-app/Dockerfile.prod
```diff
@@ -0,0 +1,30 @@
...
+
+FROM node:12-alpine As production
+
+ENV TYPEORM_ENTITIES=dist/**/*.entity.js
+ENV TYPEORM_SYNCHRONIZE=true
+ENV NODE_ENV=production
+
+WORKDIR /usr/app
+
+COPY . .
```
Clebiez, line 23: I think this line should be useless, you have just to copy your lockfile + your package json

floo51 (approved):

web-app/Dockerfile.prod
atilbian: Some other files are needed (at least those in public/ and views/) but I still get an error when I copy only these files, so it seems like others are missing. I will create an issue to reduce the production image size and I propose to merge it now.

atilbian: Related issue here : https://github.com/marmelab/Hex/issues/36

Clebiez (approved):

PR #37, opened 2022-02-21
Draw board game

atilbian:
- [x] Draw svg in Expo
- [x] Draw backbone of board
- [x] Draw board from a given game state(static)
- [x] Clean magic numbers
- [x] Extract code into a new component
- [x] ~Refactor code to a more functionnal style~ it seems that the imperative style is better for the readability here
- [x] Check how to make onPress event work for mobile and web app(try to wrap polygon into a button) =>wrapping polygon inside a button is breaking the ui(polygone aren't display anymore) so i choose to add an onClick event on polygone to make the app work in browser. I found a better fix but didn't try it : https://github.com/react-native-svg/react-native-svg/issues/1483#issuecomment-829497502
- [x] Test the app on mobile
- [x] Tests components/functions

The code to draw an hexagonal grid come from : https://gist.github.com/bencates/5b490ed79796cbd35863
The global style of the board game come from : https://hex.plus/

Related to https://trello.com/c/nxx9Sz2D/25-3-as-norbert-i-want-to-play-even-when-im-not-at-home

## Demo
On the web:
image
On mobile(horizontal scroll is allowed):
index


fzaninotto (commented):
Very clear so far. My remarks are in fact covered by the remaining items in the PR description.

mobile-app/components/Board.tsx
```diff
@@ -0,0 +1,91 @@
...
+const CELL_SIZE = 20;
+const DISTANCE_BETWEEN_HEXAGONE_PARALLEL_SIDES = Math.sqrt(3);
+
+interface BoardData {
+  gameState: GameState;
+}
+
+export default function Board(props: BoardData) {
+  return (
+    <Svg width={getApproximateSvgSize(props.gameState.board.length)} height={getApproximateSvgSize(props.gameState.board.length)}>
```
fzaninotto: `getApproximateSvgSize(props.gameState.board.length)` is repeated, extract it to a const outside of the return JSX

mobile-app/components/Board.tsx
```diff
@@ -0,0 +1,91 @@
...
+  return (col == 0 && row == 0 || col == boardSizeWithBorder - 1 && row == boardSizeWithBorder - 1);
+}
+
+function getCellType(col: number, row: number, boardSizeWithBorder: number): CellType {
+  return row == 0 || row == boardSizeWithBorder - 1 ?
+    "player2Border" :
+    col == 0 && row != boardSizeWithBorder - 1 || col == boardSizeWithBorder - 1 && row != 0 ?
+      "player1Border" :
+      "playable";
+}
```
fzaninotto: missing EOL at the end of file. This should be in your IDE settings

mobile-app/components/BorderCell.tsx
```diff
@@ -0,0 +1,21 @@
...
+  strokeColor: string;
+  playerBorder: CellType;
+}
+
+
+export default function BorderCell(props: BorderCellData) {
+  return (
+    <Polygon points={props.svgPoints} stroke={props.strokeColor} fill={getBorderColor(props.playerBorder)} />
+  )
+}
```
fzaninotto:
```suggestion
export default function BorderCell({ svgPoints, strokeColor, playerBorder }: BorderCellData) {
  return (
    <Polygon points={svgPoints} stroke={strokeColor} fill={getBorderColor(playerBorder)} />
  )
}
```

Is easier to read IMHO

mobile-app/components/Board.tsx
```diff
@@ -0,0 +1,91 @@
...
+  return points.join(' ');
+}
+
+function isCellAtTopLeftOrBottomRight(col: number, row: number, boardSizeWithBorder: number) {
+  return (col == 0 && row == 0 || col == boardSizeWithBorder - 1 && row == boardSizeWithBorder - 1);
+}
+
+function getCellType(col: number, row: number, boardSizeWithBorder: number): CellType {
+  return row == 0 || row == boardSizeWithBorder - 1 ?
+    "player2Border" :
```
fzaninotto: you should use a constant for that

mobile-app/components/BorderCell.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+function getBorderColor(cellType: CellType) {
+  return cellType == "player1Border" ? PLAYER_1_CELL_COLOR : PLAYER_2_CELL_COLOR;
```
fzaninotto: always use `=== ` unless you really want to force JS to cast type.

mobile-app/components/PlayableCell.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { DEFAULT_CELL_COLOR, PLAYER_1_CELL_COLOR, PLAYER_2_CELL_COLOR } from "../utils";
+
+interface PlayableCellData {
+  svgPoints: string;
+  strokeColor: string;
+  cellValue: Cell;
+  onCellPress: () => void;
+}
+
+export default function PlayableCell(props: PlayableCellData) {
```
fzaninotto: same remarke, use argument destructuring for shorter JSX

mobile-app/components/PlayableCell.tsx
```diff
@@ -0,0 +1,26 @@
...
+      points={props.svgPoints}
+      stroke={props.strokeColor}
+      fill={getPawnColor(props.cellValue)}
+      onClick={() => props.onCellPress()}
+      onPress={() => props.onCellPress()}
+    />)
+}
+
+function getPawnColor(cellValue: Cell) {
+  return cellValue.value == "white" ? PLAYER_1_CELL_COLOR : cellValue.value == "black" ? PLAYER_2_CELL_COLOR : DEFAULT_CELL_COLOR;
```
fzaninotto: the "white" and "black" strings should be constants. Also, use triple equal.

fzaninotto (changes requested):
Nice, but there is absolutely no test...

fzaninotto:
also, on Mobile, you must reduce the slot size so that it fits on a screen.

mobile-app/components/Board.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+const CELL_STROKE_COLOR = "white";
+const CELL_SIZE = 20;
+const DISTANCE_BETWEEN_HEXAGONE_PARALLEL_SIDES = Math.sqrt(3);
+
+interface BoardData {
+  gameState: GameState;
+}
+
+export default function Board(props: BoardData) {
```
Clebiez:
```suggestion
export default function Board({ gameState } : BoardData) {

```

same than below, it's easier to read like this

mobile-app/components/Board.tsx
```diff
@@ -0,0 +1,91 @@
...
+import PlayableCell from './PlayableCell';
+import BorderCell from './BorderCell';
+import { GameState } from '../../web-app/src/common/gameState';
+import { CellType } from "../utils";
+
+const CELL_STROKE_COLOR = "white";
+const CELL_SIZE = 20;
+const DISTANCE_BETWEEN_HEXAGONE_PARALLEL_SIDES = Math.sqrt(3);
+
+interface BoardData {
```
Clebiez: BoardData is also just a wrapper to Board's props.
I suggest to name it **BoardProps** to respect the react language

mobile-app/components/Board.tsx
```diff
@@ -0,0 +1,91 @@
...
+              onCellPress={() => onCellPress(cell.withoutBorderCoordinates)}
+            />
+            :
+            <BorderCell
+              key={index}
+              svgPoints={cell.svgPointsToDraw}
+              strokeColor={CELL_STROKE_COLOR}
+              playerBorder={cell.cellType}
+            />
+        ))
```
Clebiez: IMO destructure properties is a good way to read faster what you have inside your object and what you use inside your function. It's easier to understand.


```suggestion
        generateBoardCells(props.gameState).map(({withoutBorderCoordinates, svgPointsToDraw, cellType}, index) => (
          cellType == "playable" ?
            <PlayableCell
              key={index}
              svgPoints={svgPointsToDraw}
              strokeColor={CELL_STROKE_COLOR}
              cellValue={props.gameState.board[withoutBorderCoordinates.y][withoutBorderCoordinates.x]}
              onCellPress={() => onCellPress(withoutBorderCoordinates)}
            />
            :
            <BorderCell
              key={index}
              svgPoints={svgPointsToDraw}
              strokeColor={CELL_STROKE_COLOR}
              playerBorder={cellType}
            />
        ))

```

mobile-app/components/Board.tsx
```diff
@@ -0,0 +1,91 @@
...
+  for (let theta = 0; theta < Math.PI * 2; theta += Math.PI / 3) {
+    let pointX = x + CELL_SIZE * Math.sin(theta);
+    let pointY = y + CELL_SIZE * Math.cos(theta);
+    points.push(pointX + ',' + pointY);
+  }
+  return points.join(' ');
+}
+
+function isCellAtTopLeftOrBottomRight(col: number, row: number, boardSizeWithBorder: number) {
+  return (col == 0 && row == 0 || col == boardSizeWithBorder - 1 && row == boardSizeWithBorder - 1);
```
Clebiez: Same than below always use `===` 
Because true == 'fromage' => true
true === 'fromage' => false

mobile-app/App.tsx
```diff
@@ -1,19 +1,39 @@
...
+import { StyleSheet, View, ScrollView } from 'react-native';
+import Board from './components/Board';
+import { GameState } from '../web-app/src/common/gameState';
+
+const gamestate: GameState = {
+  turn: 'white',
+  winner: null,
+  winningPath: null,
+  board: [
+    [{ value: 'empty' }, { value: 'empty' }, { value: 'empty' }, { value: 'empty' }, { value: 'empty' }, { value: 'empty' }, { value: 'empty' }, { value: 'empty' }, { value: 'empty' }],
```
Clebiez, line 11: 'empty' should be a constant, same for both player values

mobile-app/App.tsx
```diff
@@ -1,19 +1,39 @@
-import { StatusBar } from 'expo-status-bar';
-import { StyleSheet, Text, View } from 'react-native';
+import * as React from 'react';
+import { StyleSheet, View, ScrollView } from 'react-native';
+import Board from './components/Board';
+import { GameState } from '../web-app/src/common/gameState';
```
Clebiez: Is it ok to use a component from another app ?

Consider copy paste it in your "mobile-app" folder, this folder should works in standalone

mobile-app/components/Board.tsx
Clebiez: I recommend to use prettier and activate format on save mode in your IDE

mobile-app/components/Board.tsx
```diff
@@ -0,0 +1,91 @@
...
+}
+
+function onCellPress(coordinates: { x: number, y: number }) {
+  alert(`x:${coordinates.x} y:${coordinates.y}`)
+}
+
+function generateBoardCells(gameState: GameState): { withoutBorderCoordinates: { x: number, y: number }, svgPointsToDraw: string, cellType: CellType }[] {
+  const cells = [];
+  const sizeWithBorder = gameState.board.length + 2;
+  const offset = (DISTANCE_BETWEEN_HEXAGONE_PARALLEL_SIDES * CELL_SIZE) / 2;
```
Clebiez: offset is also a constant and should be declared just below DISTANCE_BETWEEN_HEXAGONE_PARALLEL_SIDES  or CELL_SIZE

mobile-app/components/Board.tsx
```diff
@@ -0,0 +1,91 @@
...
+  for (let col = 0; col < sizeWithBorder; col += 1) {
+    for (let row = 0; row < sizeWithBorder; row += 1) {
+      if (!isCellAtTopLeftOrBottomRight(col, row, sizeWithBorder)) {
+        const rowShifter = offset * row;
+        const svgX = offset * (1 + col) * 2 + rowShifter;
+        const svgY = offset * (1 + row) * DISTANCE_BETWEEN_HEXAGONE_PARALLEL_SIDES;
+        cells.push({
+          withoutBorderCoordinates: { x: col - 1, y: row - 1 },
+          svgPointsToDraw: getSvgPoints(svgX, svgY),
+          cellType: getCellType(col, row, sizeWithBorder)
```
Clebiez: a cell object can have a type. In this case naming it "cellType" make you just repeat the words "cell".

Instead you can use cell.type which is more readable than cell.cellType.

Clebiez (changes requested):
Object and Props destructuring arethe best way to make a code readable for free.

UI 2 cents tips :
- pick colors from a palette tools and never use "red" or "black". Always use a black not totally black, a red not totally red etc...
- white border are too strong, a medium grey should be sweet !

mobile-app/components/Board.tsx
Clebiez: Oh, be careful, I code inside a comment. it could not works well !

mobile-app/App.tsx
atilbian: We are using an interface defined previously : 

```
export interface Cell {
  value: 'empty' | 'black' | 'white';
}
```

So we already have some protection with type checking and autocompletion, so does consts values are still needed here ?

mobile-app/App.tsx
atilbian: Florian gave us a solution for this(sharing all the common stuff with a local package and npm link) but I haven't tried it yet ! Does it work for you?

mobile-app/App.tsx
Clebiez: Good point for the type checking 👍 
BTW I suggest you to use enum : https://www.typescriptlang.org/docs/handbook/enums.html

mobile-app/components/PlayableCell.tsx
atilbian: Same answer than for Clement(https://github.com/marmelab/Hex/pull/37#discussion_r811962308) we are using an union type here so we have some protection against typing invalid values. Is it enough ?

mobile-app/App.tsx
atilbian: We were using enums in our first attempt, but we have been advised to use union type instead of enums for this kind of types. Now we have a lot of our codebase that's use this type so I think it's better for this run to stick with it !

mobile-app/components/Board.tsx
atilbian: I think the issue is the same as for https://github.com/marmelab/Hex/pull/37#discussion_r811943723 so we could discuss it there

atilbian:
> Nice, but there is absolutely no test...

Yes, I don't know exactly what to test (UI rendering, functions inside the components, both ?) and how to do it. Do you have good doc on how to test react native components?

mobile-app/App.tsx
Clebiez: Yep it's very complicated just for factorize `GameState` 

I think you can just copy paste `GameState` inside `mobile-app` folder and it should be good :)

Clebiez: It's just a little bit ugly to import something from another project, that's the main point here

mobile-app/App.tsx
Clebiez: You're right, after searching on google, majority of articles talks about using union types over enum :)
i'm agree with you 👍

Clebiez:
>

atilbian:
> also, on Mobile, you must reduce the slot size so that it fits on a screen.

It depends of the board size, but for a 19x19 board we need really small cells to be able to display all of these in the screen(at least on my phone) so it's very difficult to click on the desired cell.

atilbian:
> Object and Props destructuring arethe best way to make a code readable for free.
> 
> UI 2 cents tips :
> 
>     * pick colors from a palette tools and never use "red" or "black". Always use a black not totally black, a red not totally red etc...
> 
>     * white border are too strong, a medium grey should be sweet !

Thx for these tips, i will try to choose some smoother colors !

Clebiez:
> > Nice, but there is absolutely no test...
> 
> Yes, I don't know exactly what to test (UI rendering, functions inside the components, both ?) and how to do it. Do you have good doc on how to test react native components?

https://jestjs.io/fr/docs/tutorial-react-native

the first test is just to render a board without any problems, and if it works, this would means a lot 👍 

Tips: to test a function inside a component, test it without testing the component :)
It means you have to copy paste theses functions into a service folder by example and make unit test on them. It's very easy in your case because you've just made pure functions !

mobile-app/components/Board.tsx
```diff
@@ -0,0 +1,92 @@
...
+import Svg from 'react-native-svg';
+import PlayableCell from './PlayableCell';
+import BorderCell from './BorderCell';
+import { GameState } from '../../web-app/src/common/gameState';
+import { CellType } from "../utils";
+
+const CELL_STROKE_COLOR = "white";
+const CELL_SIZE = 20;
+const DISTANCE_BETWEEN_HEXAGONE_PARALLEL_SIDES = Math.sqrt(3);
+const HEXAGONE_OFFSET = DISTANCE_BETWEEN_HEXAGONE_PARALLEL_SIDES * CELL_SIZE / 2;
```
slax57:
```suggestion
const DISTANCE_BETWEEN_HEXAGON_PARALLEL_SIDES = Math.sqrt(3);
const HEXAGON_OFFSET = DISTANCE_BETWEEN_HEXAGON_PARALLEL_SIDES * CELL_SIZE / 2;

```

mobile-app/App.tsx
```diff
@@ -1,19 +1,39 @@
...
+}
 
 export default function App() {
   return (
     <View style={styles.container}>
-      <Text>Hex game for mobile</Text>
-      <StatusBar style="auto" />
+      <ScrollView horizontal>
+        <Board gameState={gamestate} />
+      </ScrollView>
```
slax57, line 28: `ScrollView horizontal` could lead to bad user experience, can't we rather reduce the size of the hexagons? Or did you just add this for really large boards?

mobile-app/components/PlayableCell.tsx
```diff
@@ -0,0 +1,26 @@
...
+    <Polygon
+      points={svgPoints}
+      stroke={strokeColor}
+      fill={getPawnColor(cellValue)}
+      onClick={() => onCellPress()}
+      onPress={() => onCellPress()}
+    />)
+}
+
+function getPawnColor(cellValue: Cell) {
```
slax57:
```suggestion
function getStoneColor(cellValue: Cell) {

```

atilbian:
> > Object and Props destructuring arethe best way to make a code readable for free.
> > UI 2 cents tips :
> > ```
> > * pick colors from a palette tools and never use "red" or "black". Always use a black not totally black, a red not totally red etc...
> > 
> > * white border are too strong, a medium grey should be sweet !
> > ```
> 
> Thx for these tips, i will try to choose some smoother colors !

Here is a first result : 
image

mobile-app/utils.ts
```diff
@@ -0,0 +1,5 @@
+export const DEFAULT_CELL_COLOR = "black";
+export const PLAYER_1_CELL_COLOR = "blue";
+export const PLAYER_2_CELL_COLOR = "red";
+
+export type CellType = "player1Border" | "player2Border" | "playable";
```
slax57:
```suggestion
export type RenderedCellType = "player1Border" | "player2Border" | "playable";

```

This is a suggestion (I also could not find a better name), because I fear this could be confusing with our already existing `Cell` interface.

web-app/src/app.service.ts
```diff
@@ -28,7 +28,7 @@ export class AppService {
     private gamesRepository: Repository<Game>,
     @InjectRepository(User)
     private usersRepository: Repository<User>,
-  ) {}
+  ) { }
```
slax57: You should avoid commiting unmodified files like this, to avoid unnecessary merge conflicts 😉

atilbian:
> > > Nice, but there is absolutely no test...
> > 
> > 
> > Yes, I don't know exactly what to test (UI rendering, functions inside the components, both ?) and how to do it. Do you have good doc on how to test react native components?
> 
> https://jestjs.io/fr/docs/tutorial-react-native
> 
> the first test is just to render a board without any problems, and if it works, this would means a lot 👍
> 
> Tips: to test a function inside a component, test it without testing the component :) It means you have to copy paste theses functions into a service folder by example and make unit test on them. It's very easy in your case because you've just made pure functions !

Ok thx i will check that !

mobile-app/App.tsx
atilbian: Check the thread here : https://github.com/marmelab/Hex/pull/37#issuecomment-1047836279 I think it will be something to think about and choose between horizontal scrolling, small cells or zooming.

mobile-app/App.tsx
atilbian: Ok will make some duplication for now and probably use the local package idea later when we will really need it.

mobile-app/package.json
```diff
@@ -12,9 +12,12 @@
   "dependencies": {
     "expo": "~44.0.0",
     "expo-status-bar": "~1.2.0",
+    "jest": "^26.6.3",
+    "jest-expo": "^44.0.1",
```
slax57: Just to be sure: you added dependencies (and real dependency, not just dev dependency, is that on purpose?) on jest and jest-expo, but I see no test file in your commits. Am I missing something?

atilbian: No, you're not. I haven't done any tests yet. I'm working on it now. For the devDepencies part, it's not my choice it's the default behaviour of `expo install jest-expo jest`(from https://docs.expo.dev/guides/testing-with-jest/).

mobile-app/components/Board.tsx
```diff
@@ -0,0 +1,91 @@
...
+}
+
+function isCellAtTopLeftOrBottomRight(col: number, row: number, boardSizeWithBorder: number) {
+  return (col === 0 && row === 0 || col === boardSizeWithBorder - 1 && row === boardSizeWithBorder - 1);
+}
+
+function getCellType(col: number, row: number, boardSizeWithBorder: number): RenderedCellType {
+  return row === 0 || row === boardSizeWithBorder - 1 ?
+    "player2Border" :
+    col === 0 && row != boardSizeWithBorder - 1 || col === boardSizeWithBorder - 1 && row != 0 ?
```
Clebiez: also consider using !== instead of !=

mobile-app/components/PlayableCell.tsx
```diff
@@ -0,0 +1,26 @@
...
+  onCellPress: () => void;
+}
+
+export default function PlayableCell({ svgPoints, strokeColor, cellValue, onCellPress }: PlayableCellProps) {
+  return (
+    <Polygon
+      points={svgPoints}
+      stroke={strokeColor}
+      fill={getStoneColor(cellValue)}
+      onClick={() => onCellPress()}
```
Clebiez:
```suggestion
      onClick={onCellPress}

```
It's the same

Clebiez (changes requested):

mobile-app/package.json
```diff
@@ -7,21 +7,31 @@
...
+    "eject": "expo eject",
+    "test": "jest",
+    "test:watch": "jest --watch",
+    "test:cov": "jest --coverage",
+    "test:update": "jest -u"
   },
   "dependencies": {
     "expo": "~44.0.0",
     "expo-status-bar": "~1.2.0",
+    "jest": "^26.6.3",
```
Clebiez: This deps should be in devDependencies

Clebiez (approved):

PR #38, opened 2022-02-22
Feature add (temporary) localplay API for use with the mobile client

slax57:
## Content

Related Trello Card: https://trello.com/c/nxx9Sz2D/25-3-as-norbert-i-want-to-play-even-when-im-not-at-home

Implements these operations:

```javascript
@Controller('localplay/games')
export class LocalplayGamesController {

    @Post('')
    initNewGameState(
        @Body() params: { size?: number }
    ): GameState {
        return initNewGameState(params.size);
    }

    @Put('')
    updateGameState(
        @Body() params: { previousGameState: GameState, nextMove: Coordinates }
    ): GameState {
        return updateGameState(params.previousGameState, params.nextMove);
    }

}
```

## Checklist

- [x] Fix (some of the) semantic of the existing REST operations
- [x] Refactor web-app code into modules (games and users)
- [x] Add a new module (containing a controller) for the temporary localplay API


web-app/src/app.module.ts
```diff
@@ -1,22 +1,15 @@
...
-    TypeOrmModule.forRootAsync({
-      useFactory: async () =>
-        Object.assign(await getConnectionOptions(), {
-          autoLoadEntities: true,
-        }),
-    }),
-    TypeOrmModule.forFeature([Game, User]),
+    GamesModule,
+    UsersModule,
+    GamesModule,
```
atilbian: LocalplayGamesModule instead ?

web-app/src/localplay/games/localplay.games.controller.ts
```diff
@@ -0,0 +1,22 @@
+import { Body, Controller, Post, Put } from '@nestjs/common';
```
atilbian: Why do you choose to not set localplay files at the top of the localplay directory ?

web-app/src/localplay/games/localplay.games.controller.ts
```diff
@@ -0,0 +1,22 @@
...
+import { GameState, initNewGameState, updateGameState } from '../../common/gameState';
+import { Coordinates } from '../../common/utils';
+
+@Controller('localplay/games')
```
atilbian: What about adding an "api" prefix into rest api endpoint for non ssr controllers ?

atilbian:
The mix of new nestjs modules and cli/commons folders is getting weirder. I know we'll pull them out later but we could create a "module" subfolder to group them all together for now ?

web-app/src/common/utils.ts
```diff
@@ -1,4 +1,4 @@
-import { Game } from 'src/entities/game.entity';
```
atilbian, line 1: I know this was done in a previous merge and I didn't see it then but I think all the entity related stuff doesn't belong here because it's very specific to the web application

web-app/src/localplay/games/localplay.games.controller.ts
slax57: Since the REST route I chose is `'localplay/games'` I figured this would be more coherent. Do you disagree?

web-app/src/localplay/games/localplay.games.controller.ts
slax57: good idea 🙂

slax57:
> The mix of new nestjs modules and cli/commons folders is getting weirder. I know we'll pull them out later but we could create a "module" subfolder to group them all together for now ?

I prefer keeping the ability to use nest CLI to generate modules for us, and I don't know if we can configure nest to create modules inside a 'module' folder...

web-app/src/games/games.controller.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    @Get(':id')
+    @Render('game')
+    async getGame(@Param('id') id: number, @Req() req: Request): Promise<GameAndStatus> {
+        const game = await this.gameService.findGameById(id);
+        return this.gameService.getGameAndStatus(game, req.sessionID);
+    }
+
+    @Post(':id')
+    @Render('game')
```
Clebiez, line 48: I'm not sure but if it's an API, why did you use Render annotation ?

web-app/src/games/games.service.spec.ts
```diff
@@ -39,13 +39,13 @@ const mockUsersRepository = {
   },
 };
 
-describe('AppService', () => {
-  let appService: AppService;
+describe('GameService', () => {
+  let gameService: GamesService;
```
Clebiez, line 58: Your gameService is a variable which will mutate between different tests.

It's a bad practice and you should avoid that.

Each `it` function should be completely independant, obviously you can use "beforeEach" and "afterEach" but you must never be able to grab a variable which is updated from another test.

If you have a test with no call to `updateGameState`, but you assert something, this assert pass but the feature doesn't work because the board that you expected is wrong.
Also you can't be certified that every tests runs will run your test in the same order.
Prefer setting the game in a beforeEach by calling `getBoardStateFromFile` every time.

atilbian:
> > The mix of new nestjs modules and cli/commons folders is getting weirder. I know we'll pull them out later but we could create a "module" subfolder to group them all together for now ?
> 
> I prefer keeping the ability to use nest CLI to generate modules for us, and I don't know if we can configure nest to create modules inside a 'module' folder...

Ok so we can stay with this for now.

web-app/src/localplay/games/localplay.games.controller.ts
atilbian: Since this is probably a temporary endpoint I think we can leave it !

web-app/src/games/games.controller.ts
atilbian: The rest api is in the localplay module. Here it's our SSR module so we still need to target a template to render.

web-app/src/games/games.service.spec.ts
slax57: Opened [this issue](https://github.com/marmelab/Hex/issues/45) for this topic

Issue #36, opened 2022-02-21
Optimize the production image size


atilbian:
https://github.com/marmelab/Hex/pull/34#discussion_r811270322
- [ ] Reduce image size by copying only required files

PR #39, opened 2022-02-22
Feature deploy to AWS EC2 instead of EBS 

slax57:
Edit github workflows to deploy to an EC2 instance instead of the full EBS (hoping future deployments will be more stable)

New deployment address: http://ec2-54-86-88-53.compute-1.amazonaws.com/

- [x] Create the EC2 instance
- [x] Add secrets to Github
- [x] Edit the workflow's yaml file
- [x] Confirm public access to the service
- [x] Change branch name to 'main' in yaml file


.github/workflows/deployment-ec2.yml
```diff
@@ -0,0 +1,27 @@
...
+
+      - name: Deploy in EC2
+        env:
+            PRIVATE_KEY: ${{ secrets.AWS_EC2_PRIVATE_KEY  }}
+            HOSTNAME : ${{ secrets.AWS_EC2_PUBLIC_IP  }}
+            USER_NAME : ${{ secrets.AWS_EC2_USERNAME  }}
+            
+        run: |
+          echo "$PRIVATE_KEY" > private_key && chmod 600 private_key
+          rsync --delete -r -e "ssh -o StrictHostKeyChecking=no -i private_key" ./ ${USER_NAME}@${HOSTNAME}:~/Hex
```
atilbian: We could be more selective about what we send to ec2(e.g exclude mobile-app/ .github folders) but I don't think it's a top priority(as for https://github.com/marmelab/Hex/issues/36)

.github/workflows/deployment-ec2.yml
```diff
@@ -0,0 +1,27 @@
+name: Deploy on aws EC2
```
atilbian: I think it's time to rename the file to be explicit that this script is for deploying the web server only and not the mobile application. What do you think about "deploy-web-ec2.yml" ?

slax57: I'll go for "deploy-web-app-ec2.yml", thanks 🙂

.github/workflows/deploy-web-app-ec2.yml
```diff
@@ -0,0 +1,27 @@
...
+
+      - name: Deploy in EC2
+        env:
+            PRIVATE_KEY: ${{ secrets.AWS_EC2_PRIVATE_KEY  }}
+            HOSTNAME : ${{ secrets.AWS_EC2_PUBLIC_IP  }}
+            USER_NAME : ${{ secrets.AWS_EC2_USERNAME  }}
+            
+        run: |
+          echo "$PRIVATE_KEY" > private_key && chmod 600 private_key
+          rsync --delete -r -e "ssh -o StrictHostKeyChecking=no -i private_key" --exclude="mobile-app/*" ./ ${USER_NAME}@${HOSTNAME}:~/Hex
```
Clebiez, line 23: Add a --filter param to use gitignore, when you will run test on CI, you will have to install dependencies and the node_modules folder will be rsynced by this step.

PR #43, opened 2022-02-23
Create rest api endpoints to play online

atilbian:
- [x] Create endpoint to create a new game
 - [x] Create endpoint to retrieve an existing game
 - [x] Create endpoint to update an existing game
 - [x] Fix broken tests
 - [x] Throw 404 not found when requested game doesn't exist

Postponed :
 - Check if other specific errors should be triggered (Internal server error-500 is the default error triggered when an exception occurs)
 -  Add tests on controllers(api and ssr) (the service layer is already tested because we are reusing the existing one)

Related to https://trello.com/c/xDXP1pon/28-2-as-norbert-i-want-to-play-with-philippe-even-when-were-not-together

## Demo
demo-play-online

### DIFF
--- a/web-app/src/games/api.games.controller.ts
+++ b/web-app/src/games/api.games.controller.ts
@@ -8,12 +8,13 @@ import {
   Req,
   UseGuards,
   NotFoundException,
+  Request
 } from '@nestjs/common';
-import { Request } from 'express';
 import { Coordinates } from '../common/utils';
 import { GamesService, GameAndStatus } from './games.service';
 import { Game } from './game.entity';
 import { AuthGuard } from '@nestjs/passport';
+import { UserDataInjectedInRequestAfterAuth } from 'src/auth/auth.service';
 
 @UseGuards(AuthGuard('jwt'))
 @Controller('api/games')
@@ -23,39 +24,40 @@ export class ApiGamesController {
   @Post('')
   async create(
     @Body() params: { size?: number },
-    @Req() req: Request
+    @Request() req: { user: UserDataInjectedInRequestAfterAuth }
   ): Promise<Game> {
-    return await this.gamesService.createNewGame(params.size, req.user["username"]);
+    return await this.gamesService.createNewGame(params.size, req.user.username);
   }
 
   @Get(':id')
   async findOne(
     @Param('id') id: number,
-    @Req() req: Request
+    @Request() req: { user: UserDataInjectedInRequestAfterAuth }
   ): Promise<GameAndStatus> {
     const game = await this.gamesService.findGameById(id);
     if (game === undefined) throw new NotFoundException();
-    return this.gamesService.getGameAndStatus(game, req.user["username"]);
+    return this.gamesService.getGameAndStatus(game, req.user.username);
   }
 
   @Put(':id')
   async update(
     @Param('id') id: number,
     @Body() params: { nextMove: Coordinates },
-    @Req() req: Request
+    @Request() req: { user: UserDataInjectedInRequestAfterAuth }
   ): Promise<GameAndStatus> {
     let game = await this.gamesService.findGameById(id);
     if (game === undefined) throw new NotFoundException();
     game = await this.gamesService.updateGameState(game, params.nextMove);
-    return this.gamesService.getGameAndStatus(game, req.user["username"]);
+    return this.gamesService.getGameAndStatus(game, req.user.username);
   }
 
   @Get(':id/join')
   async joinGame(
     @Param('id') id: number,
-    @Req() req: Request) {
+    @Request() req: { user: UserDataInjectedInRequestAfterAuth }
+  ): Promise<Game> {
     const game = await this.gamesService.findGameById(id);
     if (game === undefined) throw new NotFoundException();
-    await this.gamesService.joinGame(id, req.user["username"]);
+    return await this.gamesService.joinGame(id, req.user.username);
   }
 }
--- a/web-app/src/games/api.games.controller.ts
+++ b/web-app/src/games/api.games.controller.ts
@@ -46,12 +46,7 @@ export class ApiGamesController {
   ): Promise<GameAndStatus> {
     let game = await this.gamesService.findGameById(id);
     if (game === undefined) throw new NotFoundException();
-    if (params.nextMove.x && params.nextMove.y) {
-      game = await this.gamesService.updateGameState(game, {
-        x: params.nextMove.x,
-        y: params.nextMove.y,
-      });
-    }
+    game = await this.gamesService.updateGameState(game, params.nextMove);
     return this.gamesService.getGameAndStatus(game, req.user["username"]);
   }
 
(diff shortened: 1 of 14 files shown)


web-app/src/auth/auth.service.ts
```diff
@@ -0,0 +1,37 @@
...
+    ) { }
+
+    async checkCredentials({ login }): string {
+        const user = await this.userRepository.findOne({ where:
+            { username: login }
+        });
+        if (user === undefined) {
+            throw new Error('unknown user')
+        }
+
```
Clebiez: Your IDE seems to not be correctly configured with prettier or eslint, you should take 5 min to configure it correctly :)

web-app/src/api/games/auth.middleware.ts
```diff
@@ -0,0 +1,10 @@
+import { Injectable, NestMiddleware } from '@nestjs/common';
+import { Request, Response, NextFunction } from 'express';
+
+@Injectable()
+export class AuthMiddleware implements NestMiddleware {
+  use(req: Request, res: Response, next: NextFunction) {
+    console.log('Request...');
```
Clebiez: console spotted

Clebiez (changes requested):

web-app/src/games/api.games.controller.ts
```diff
@@ -0,0 +1,64 @@
...
+    @Body() params: { nextMove: Coordinates },
+    @Req() req: Request
+  ): Promise<GameAndStatus> {
+    let game = await this.gamesService.findGameById(id);
+    if (game === undefined) throw new NotFoundException();
+    if (params.nextMove.x && params.nextMove.y) {
+      game = await this.gamesService.updateGameState(game, {
+        x: params.nextMove.x,
+        y: params.nextMove.y,
+      });
```
slax57:
```suggestion
      game = await this.gamesService.updateGameState(game, params.nextMove);
```
wouldn't this work?

web-app/src/games/api.games.controller.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+    return this.gamesService.getGameAndStatus(game, req.user["username"]);
+  }
+
+  @Get(':id/join')
+  async joinGame(
+    @Param('id') id: number,
+    @Req() req: Request) {
+    await this.gamesService.joinGame(id, req.user["username"]);
```
slax57: It would be great not to have to copy `req.user["username"]` everytime, but I don't know how we could avoid it... If s.o. knows, I'd be curious 🙂

web-app/src/games/games.service.ts
```diff
@@ -53,60 +52,51 @@ export class GamesService {
...
-                throw Error(`Player2 (sessionId: ${player2SessionId}) cannot be the same as player1 (sessionId: ${game.player1.lastSessionId})`);
+                throw Error(`Player2 cannot be the same user than player1`);
             }
         } else {
             throw Error(`Game ${gameId} already has a player2 (username: ${game.player2.username})`);
         }
     }
 
-    getGameAndStatus(game: Game, playerSessionId: string): GameAndStatus {
+    getGameAndStatus(game: Game, playerName: string): GameAndStatus {
```
slax57, line 91:
```suggestion
    getGameAndStatus(game: Game, playerUserName: string): GameAndStatus {
```
Naming would be more consistant this way 🙂

web-app/src/games/api.games.controller.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  @Get(':id')
+  async findOne(
+    @Param('id') id: number,
+    @Req() req: Request
+  ): Promise<GameAndStatus> {
+    const game = await this.gamesService.findGameById(id);
+    if (game === undefined) throw new NotFoundException();
+    return this.gamesService.getGameAndStatus(game, req.user["username"]);
+  }
```
slax57: It would be nice to filter out some of the players data, such as the password hash which is still present in the response 😅 

(here is the output with POST, but we have the same prb with GET)

2022-02-24 17_01_44-Clipboard

web-app/src/games/api.games.controller.ts
atilbian: The best I can do for now is to use the new type coming from https://github.com/marmelab/Hex/pull/47 !
```ts
@Get(':id/join')
  async joinGame(
    @Param('id') id: number,
    @Request() req: { user: UserDataInjectedInRequestAfterAuth }
  ) {
    const game = await this.gamesService.findGameById(id);
    if (game === undefined) throw new NotFoundException();
    await this.gamesService.joinGame(id, req.user.username);
  }
```

web-app/src/games/api.games.controller.ts
atilbian: Yes, I think there is something missing between the entity layer and the service layer to transform the data (and hide sensitive data for example).

web-app/src/games/games.service.ts
```diff
@@ -53,60 +51,51 @@ export class GamesService {
...
         if (!game.player2) {
-            if (player2SessionId !== game.player1.lastSessionId) {
-                const player2 = this.usersRepository.create({
-                    lastSessionId: player2SessionId,
-                    username: player2SessionId
-                });
-                game.player2 = await this.usersRepository.save(player2);
+            if (player2Username !== game.player1.username) {
+                const player2 = await this.usersService.getOrCreate(player2Username);
+                game.player2 = player2;
```
Clebiez:
```suggestion
                game.player2 = await this.usersService.getOrCreate(player2Username);
```

web-app/src/games/games.service.ts
```diff
@@ -53,60 +51,51 @@ export class GamesService {
...
-        const player1 = this.usersRepository.create({
-            lastSessionId: player1SessionId,
-            username: player1SessionId
-        });
+    async createNewGameFromFile(player1Username: string): Promise<Game> {
+        const player1 = await this.usersService.getOrCreate(player1Username);
         const game = this.gamesRepository.create({
             state: this.getBoardStateFromFile(),
-            player1: await this.usersRepository.save(player1),
+            player1: player1,
```
Clebiez:
```suggestion
            player1,
```

web-app/src/games/games.service.ts
```diff
@@ -53,60 +51,51 @@ export class GamesService {
...
         }
-        const player1 = this.usersRepository.create({
-            lastSessionId: player1SessionId,
-            username: player1SessionId
-        });
+        const player1 = await this.usersService.getOrCreate(player1Username);
         const game = this.gamesRepository.create({
             state: initNewGameState(size),
-            player1: await this.usersRepository.save(player1),
+            player1: player1,
```
Clebiez:
```suggestion
            player1,
```

Clebiez (changes requested):

Clebiez (approved):

Clebiez (approved):

PR #42, opened 2022-02-23
Add navigation backbone

atilbian:
- [x] Init a navigation system
- [x] Add home page
- [x] Display buttons to create a game locally or online or join a game
- [x] Replace snapshot tests (instead of maintaining them) with more robust tests.
- [x] Use typescript for new tests
- [ ] (For later ?) Clean last warning about the returned type of renderer.create call 

Related to https://trello.com/c/xDXP1pon/28-2-as-norbert-i-want-to-play-with-philippe-even-when-were-not-together

## Demo
demo-home-page-and-nav

PR #44, opened 2022-02-23
Mobile app: Call Hex API to create and update board for a local play

slax57:
## Content

Trello card: https://trello.com/c/nxx9Sz2D/25-3-as-norbert-i-want-to-play-even-when-im-not-at-home

- [x] Create a LocalPlayService, responsible for calling the Hex API
- [x] Use env vars to configure API base URL
- [x] Add a state to the 'App' component to hold and update the game state
- [x] Init game state and handle clicks thanks to the LocalPlayService
- [x] Enable CORS on the web app --> moved to https://github.com/marmelab/Hex/pull/46
- [x] Update Readme
- [x] Test calling API deployed on AWS
- ~[ ] Update / add automated tests - ABORTED~
- [x] Rebase with main
- [x] Prepare app for packaging (tested with APK)

## Demo
hex-mobile-localplay-api


mobile-app/App.tsx
```diff
@@ -1,30 +1,32 @@
...
+};
 
 export default function App() {
+  const [gameState, setGameState] = useState<GameState>(gameStateInitialValue);
+
+  React.useEffect(() => {
+    initNewGameState(9).then((data) => setGameState(data));
+  }, []);
+
+  const setGameStateWrapper = (gameState: GameState) => {
```
fzaninotto: why do you need this wrapper ?

mobile-app/api-client/LocalplayService.ts
```diff
@@ -0,0 +1,38 @@
+import { GameState, Coordinates } from "../utils";
+import { HEX_API_URL } from "@env"
+
+const LOCALPLAY_GAMES_ENDPOINT = "/localplay/games";
+
+export async function initNewGameState(size?: number): Promise<GameState> {
+  console.debug(`Call initNewGameState() with size=${size}`);
```
fzaninotto: console.log spotted

mobile-app/api-client/LocalplayService.ts
```diff
@@ -0,0 +1,38 @@
...
+  console.debug(`Full URL: ${HEX_API_URL + LOCALPLAY_GAMES_ENDPOINT}`);
+  const response = await fetch(HEX_API_URL + LOCALPLAY_GAMES_ENDPOINT, {
```
fzaninotto: we don't usually use `+`to concatenate strings. Prefer template literals with backticks.

```js
await fetch(`${HEX_API_URL}${LOCALPLAY_GAMES_ENDPOINT}`)
```

README.md
```diff
@@ -0,0 +1,18 @@
...
+## Running the app
+
+### With Docker Compose
+
+```bash
+$ docker-compose -f docker-compose.base.yaml -f docker-compose.dev.yaml up -d
+
+# Or if you want to use the prod env
+$ docker-compose -f docker-compose.base.yaml -f docker-compose.prod.yaml up -d
+```
```
atilbian: Nice !

mobile-app/.gitignore
```diff
@@ -8,4 +8,5 @@ npm-debug.*
 *.key
 *.mobileprovision
 *.orig.*
-web-build/
\ No newline at end of file
+web-build/
+.env
```
atilbian, line 12: I think that's env files are already ignored in the gitignore at the top of the repo

mobile-app/api-client/LocalplayService.ts
```diff
@@ -0,0 +1,38 @@
+import { GameState, Coordinates } from "../utils";
```
atilbian: To be consistant we need to use lowerCamelCase for this file(and UpperCamelCase for react components)

web-app/src/main.ts
```diff
@@ -18,6 +18,7 @@ async function bootstrap() {
       saveUninitialized: true,
     }),
   );
+  app.enableCors();
```
atilbian: We could be more selective and allow only our local Expo web app to consume our API

mobile-app/App.tsx
```diff
@@ -1,30 +1,32 @@
...
-}
+const gameStateInitialValue: GameState = {
+  board: [],
+  turn: "white",
+  winner: null, 
+  winningPath: null
+};
 
 export default function App() {
+  const [gameState, setGameState] = useState<GameState>(gameStateInitialValue);
```
atilbian: I don't understand why a default value is needed, because we want to render the component after the result of our api call. But it's seems to be the way to go with React hooks.

mobile-app/.env.prod.dist
```diff
@@ -0,0 +1 @@
+HEX_API_URL=http://myhostname/api
```
atilbian: Where will this file be used? It's ok for the env one(in the docker-compose.dev) but I'm not sure for this one

web-app/src/main.ts
slax57: Why not? Its destiny is to become a public API one day, isn't it? 😁

mobile-app/App.tsx
slax57: Today the problem is, this is the home page of our app, and we can't afford to wait on the API to answer something to display our home page. But once https://github.com/marmelab/Hex/pull/42 will have been merged, this will no longer be a problem and we'll be able to get rid of this initial value 🙂

mobile-app/.env.prod.dist
slax57: When we'll package our app as and APK for instance, we well need a `.env` file containing the URL for the API to use in production. This is where this template comes at play, it gives the template for this `.env` file.

But since our API is supposed to be open and public, I guess I could have filled it with the true URL right away... What do you think?

mobile-app/.gitignore
slax57: Didn't work for me in VSCode, because I opened the folder `mobile-app` as a project (and not the root of the repo). That's why I think it makes sense to add this line here.

web-app/src/main.ts
atilbian: Yep, one day but not for now :p

mobile-app/App.tsx
atilbian: Ok, I get it thx !

mobile-app/.env.prod.dist
atilbian: I don't know how apk generation works but for the moment this file is not used anywhere so for me there is no need to keep it !

mobile-app/.env.dev.dist
```diff
@@ -0,0 +1 @@
+HEX_API_URL=http://localhost:3000/api
```
atilbian, line 1: I think the .env.dev file need to be injected into the expo service in the docker-compose.dev file !

mobile-app/api-client/LocalplayService.ts
```diff
@@ -0,0 +1,38 @@
...
+  nextMove: Coordinates): Promise<GameState> {
+  const response = await fetch(HEX_API_URL + LOCALPLAY_GAMES_ENDPOINT, {
+    method: 'PUT',
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json'
+    },
+    body: JSON.stringify({
+      previousGameState: previousGameState,
+      nextMove: nextMove
```
Clebiez:
```suggestion
      previousGameState,
      nextMove
```

In a object, when a value has a same name than the key, we can do this, and it's cool

mobile-app/components/Board/Board.tsx
```diff
@@ -4,12 +4,18 @@ import PlayableCell from '../PlayableCell/PlayableCell';
...
 
 interface BoardProps {
   gameState: GameState;
+  setGameState: (gameState: GameState) => void;
 }
 
-export default function Board({ gameState }: BoardProps) {
+export default function Board({ gameState, setGameState }: BoardProps) {
+  const onCellPress = (coordinates: Coordinates) => {
+    updateGameState(gameState, coordinates).then((data) => setGameState(data));
```
Clebiez:
```suggestion
    updateGameState(gameState, coordinates).then(setGameState);

```

Clebiez (changes requested):

web-app/src/main.ts
slax57: As seen together, this was finally done with PR https://github.com/marmelab/Hex/pull/46
For now I left this configuration (allow all origins) because I did not have enough time to test the config allowing only one origin.

mobile-app/.env.prod.dist
slax57: I've updated its content to

```
HEX_API_URL=http://ec2-54-86-88-53.compute-1.amazonaws.com/api
```

For me it's a good thing to have both a local development config example, and a production environment example. What's more, if (one day) we get to automatize build of our mobile app, then in the script we may write something like `cp .env.prod.dist .env` to have this file automatically created.

Give me a call if you still disagree, so that we can resolve this together 🙂

mobile-app/.env.dev.dist
slax57: Actually the file needed is called `.env` and not `.env.dev` (RTFM buddy!! 🤣 )

And I can confirm it's embedded correctly both with expo build and docker-compose 🙂

mobile-app/src/api-client/localplayService.ts
```diff
@@ -0,0 +1,36 @@
...
+    method: 'POST',
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json'
+    },
+    body: JSON.stringify({
+      size,
+    })
+  });
+  return await response.json();
```
Clebiez: Duplicated code between all of your calls, you can factorize it

mobile-app/src/components/Board/Board.tsx
```diff
@@ -34,8 +40,4 @@ export default function Board({ gameState }: BoardProps) {
       }
     </Svg >
   );
-}
-
-function onCellPress(coordinates: Coordinates) {
-  alert(`x:${coordinates.x} y:${coordinates.y}`)
-}
+}
```
Clebiez: No newline at end of file, you need a well configured prettier :)

Clebiez (changes requested):

PR #47, opened 2022-02-24
Add create and login endpoints

atilbian:
- [x] Add password on user entity
- [x] Add endpoint to create new user
- [x] Add endpoint get an JWT token based on username/password
- [x] Add endpoint to test the authorization mechanism

Security related issues :
- [x] Encrypt password in database
- [x] Don't store the jwtConstants in repo

Related to https://trello.com/c/xDXP1pon/28-2-as-norbert-i-want-to-play-with-philippe-even-when-were-not-together

## DEMO
demo-create-account


Makefile
```diff
@@ -0,0 +1,8 @@
+init:
```
arimet, line 1: Next time you can also look at making a make install for the dependencies :)

web-app/.env.prod.dist
```diff
@@ -2,4 +2,5 @@ TYPEORM_CONNECTION=postgres
 TYPEORM_HOST=postgres
 TYPEORM_USERNAME=postgres
 TYPEORM_PASSWORD=postgres
-TYPEORM_DATABASE=test
\ No newline at end of file
+TYPEORM_DATABASE=test
+JWT_SECRET=aaabbbccc
```
arimet: Are you sure you want to push confidential information on the depot?

web-app/.env.dev.dist
```diff
@@ -2,4 +2,5 @@ TYPEORM_CONNECTION=postgres
 TYPEORM_HOST=postgres
 TYPEORM_USERNAME=postgres
 TYPEORM_PASSWORD=postgres
-TYPEORM_DATABASE=test
\ No newline at end of file
+TYPEORM_DATABASE=test
+JWT_SECRET=aaabbbccc
```
Clebiez:
```suggestion
JWT_SECRET=!ChangeMe!
```

We generally use this as default value to not forget to change it :)

web-app/.env.prod.dist
Clebiez: JWT secret must be setted as env var and injected by using your CI during your deploy, you can't use it a clear value in production.

web-app/src/users/api.users.controller.ts
```diff
@@ -0,0 +1,21 @@
...
+
+@Controller('api/users')
+export class ApiUsersController {
+  constructor(private readonly usersService: UsersService) { }
+
+  @Post('')
+  async create(
+    @Req() req: Request
+  ): Promise<User> {
+    const { username, password } = req.body;
```
arimet: Use @Body decorator : https://docs.nestjs.com/controllers

web-app/src/auth/auth.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+import { AuthService } from './auth.service';
+
+@Controller('api/auth')
+export class AuthController {
+
+  constructor(private authService: AuthService) { }
+
+  @UseGuards(AuthGuard('local'))
+  @Post('/login')
+  async login(@Request() req) {
```
arimet: Can't you be more specific about what is expected in user?

web-app/src/auth/auth.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+    return this.authService.login(req.user);
+  }
```
slax57: You can give info about the type expected for you param `user` like I did in the games controller:

```ts
async createNewGame(@Body() gameParams: { size?: number, fromFile?: boolean }, @Req() req: Request) {
  ...
}
```

web-app/src/auth/jwt.strategy.ts
```diff
@@ -0,0 +1,20 @@
...
+import { PassportStrategy } from '@nestjs/passport';
+import { Injectable } from '@nestjs/common';
+import { jwtConstants } from './auth.constant';
+
+@Injectable()
+export class JwtStrategy extends PassportStrategy(Strategy) {
+  constructor() {
+    super({
+      jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
+      ignoreExpiration: true,
```
slax57, line 13: Are we supposed to keep that line?

web-app/src/users/api.users.controller.ts
slax57: Same here, you should type the expected body

web-app/src/users/users.module.ts
```diff
@@ -1,5 +1,23 @@
...
 @Module({
+  imports: [
+    TypeOrmModule.forRootAsync({
+      useFactory: async () =>
+        Object.assign(await getConnectionOptions(), {
+          autoLoadEntities: true,
+          keepConnectionAlive: true,
+        }),
+    }),
+    TypeOrmModule.forFeature([User]),
```
slax57: I'm surprised you have to copy all this config code here as well.

Wouldn't the last line (`TypeOrmModule.forFeature([User]),`) be sufficient maybe?

web-app/.env.prod.dist
slax57: The philosophy of this file is to give an example value for each param, as we could/should have them in production. But the real values are created from a github secret, don't worry 😉 .

As I side note, on second thought I also agree that having this file is arguable. Maybe the `.env.dev.dist` file is enough to serve as a template...

Makefile
atilbian: Currently it's all included into the "start" cmd because npm depencies are installed and builded during the image construction !

web-app/.env.prod.dist
atilbian: Yes, we should probably delete the prod one !

web-app/src/auth/jwt.strategy.ts
atilbian: Not in a real production environnement but for now it can make things a little easier for us !

web-app/src/users/users.module.ts
atilbian: I will try without it !

web-app/src/auth/auth.controller.ts
atilbian: I don't know how to do it, because the user object is injected from the guard mechanism and don't know how to retrieve the returned type...

web-app/src/auth/auth.controller.ts
atilbian: Same comment than Anthony here : https://github.com/marmelab/Hex/pull/47#discussion_r813921727

web-app/src/auth/auth.controller.ts
atilbian: I propose to leave it as is for the moment!

web-app/src/users/users.service.ts
```diff
@@ -0,0 +1,33 @@
...
+  constructor(
+    @InjectRepository(User)
+    private usersRepository: Repository<User>,
+  ) { }
+
+  private async hashPassword(password: string) {
+    return await hash(password, 10);
+  }
+
+  async findOne(username: string, password: string): Promise<User | undefined> {
```
Clebiez: wrong naming, it should be login because you compare password too

web-app/src/auth/auth.service.ts
```diff
@@ -0,0 +1,28 @@
...
+import { UsersService } from 'src/users/users.service';
+
+@Injectable()
+export class AuthService {
+  constructor(
+    private usersService: UsersService,
+    private jwtService: JwtService
+  ) { }
+
+  async validateUser(username: string, pass: string): Promise<any> {
```
mchaffotte: avoid any

Clebiez (changes requested):

web-app/src/auth/auth.service.ts
```diff
@@ -0,0 +1,28 @@
...
+    const user = await this.usersService.findOne(username, pass);
+    if (user) {
+      const { password, ...result } = user;
+      return result;
+    }
+    return null;
+  }
+
+  async login(user: any) {
```
mchaffotte: any! and return type is missing

web-app/src/auth/jwt.strategy.ts
```diff
@@ -0,0 +1,20 @@
...
+      secretOrKey: jwtConstants.secret,
+    });
+  }
+
+  async validate(payload: any) {
```
mchaffotte: any and return type is missing

web-app/src/auth/local.strategy.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Injectable, UnauthorizedException } from '@nestjs/common';
+import { AuthService } from './auth.service';
+
+@Injectable()
+export class LocalStrategy extends PassportStrategy(Strategy) {
+  constructor(private authService: AuthService) {
+    super();
+  }
+
+  async validate(username: string, password: string): Promise<any> {
```
mchaffotte: avoid any

web-app/src/users/api.users.controller.ts
```diff
@@ -0,0 +1,19 @@
...
+import { User } from './user.entity';
+import { UsersService } from './users.service';
+
+@Controller('api/users')
+export class ApiUsersController {
+  constructor(private readonly usersService: UsersService) { }
+
+  @Post('')
+  async create(
+    @Body() params: { username: string, password: string }
```
mchaffotte: extract your type

web-app/src/users/users.service.ts
```diff
@@ -0,0 +1,33 @@
...
+
+@Injectable()
+export class UsersService {
+  constructor(
+    @InjectRepository(User)
+    private usersRepository: Repository<User>,
+  ) { }
+
+  private async hashPassword(password: string) {
+    return await hash(password, 10);
```
mchaffotte: why do you await a promise whereas you do nothing with?

mchaffotte: missing return type

web-app/src/users/users.service.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  async findOne(username: string, password: string): Promise<User | undefined> {
+    const user = await this.usersRepository.findOne({
+      where: { username: username }
+    });
+    return await compare(password, user.password) ? user : undefined;
```
mchaffotte: why do you await a promise whereas you do nothing with?

web-app/src/users/users.service.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  async create(username: string, password: string): Promise<User> {
+    const user = this.usersRepository.create({
+      username: username,
+      password: await this.hashPassword(password)
+    });
+    return await this.usersRepository.save(user)
```
mchaffotte: why do you await a promise whereas you do nothing with?

mchaffotte (changes requested):

PR #57, opened 2022-02-28
Fix makefile init mobile app

slax57:
## Content

- [x] Fix missing env file for mobile-app in makefile init
- [x] Replace docker compose commands by make in global readme file


Makefile
```diff
@@ -2,10 +2,11 @@ init:
 	@cp ./web-app/.env.dist ./web-app/.env
 	@cp ./.docker/pgadmin/.env.pgadmin.dist ./.docker/pgadmin/.env.pgadmin
 	@cp ./.docker/db/.env.db.dist ./.docker/db/.env.db
+	@cp ./mobile-app/.env.dev.dist ./mobile-app/.env
```
atilbian: To be consistent with other we could remove the .dev/.prod to keep only one file !

PR #58, opened 2022-02-28
Improve player's turn and win message

slax57:
## Content

- [x] Edit message while in remote play to clarify if it's my turn to play or not
- [x] Improved win message to make it clearer if I won the game or lost it
- [x] Update automated tests
- [x] Use fixed height for messages to avoid screen blinking

Related to: https://trello.com/c/rUzI9XZe/69-05-improve-ui-for-current-player-in-distant-mode

### Win
hex-mobile-remote-win-fixed-height

### Loose
hex-mobile-remote-loose-fixed-height

### DIFF
--- a/mobile-app/src/components/Board/Board.tsx
+++ b/mobile-app/src/components/Board/Board.tsx
@@ -26,7 +26,7 @@ export default function Board({
   const svgSize = getApproximateSvgSize(gameState.board.length);
   return gameState.winner ? (
     <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
-      {getWinMsgElt(gameState)}
+      <View style={{ height: '100px' }}>{getWinMsgElt(gameState)}</View>
       <Svg width={svgSize} height={svgSize}>
         {generateBoardCells(gameState).map(
           ({ withoutBorderCoordinates, svgPointsToDraw, type }, index) =>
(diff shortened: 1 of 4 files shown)


mobile-app/src/components/Board/Board.tsx
```diff
@@ -7,23 +7,26 @@ import {
...
+  getPlayersTurnElt,
+  getWinMsgElt,
+}: BoardProps) {
   const svgSize = getApproximateSvgSize(gameState.board.length);
   return gameState.winner ? (
     <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
-      <Text style={{ fontSize: 32 }}>
-        Player {mapStoneColorToPlayerName(gameState.winner)} has won !
-      </Text>
+      <View style={{ height: '100px' }}>{getWinMsgElt(gameState)}</View>
```
atilbian: I think we can now extract that redundant part and wrap children inside a React Fragment
```tsx
<View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
```

djhi:
A little hover effect would make it easier to use IMO

mobile-app/src/components/Board/Board.tsx
slax57: done

atilbian:
> A little hover effect would make it easier to use IMO

Yes i agree it could be usefull but for dev only, because this app is build as a mobile app only. So i don't know if it's a good idea to add features specifics to web app. I merge this PR as it is, and we can discuss about this later.

PR #65, opened 2022-03-01
Add games panel and feed it with ou game entities

atilbian:
- [x] Test Game list component => Not required but i have done one test just to be used as an example
- [x] Test our funcs implemented into data provider ? => Yes because it's the part of our app with the most of specific code but doesn't seems usefull for now as the current data provider is very simple.
- [x] Add deleteMany func in data provider
- [x] Update getList to allow with filtering, ordering and pagination
- [x] Fix batch delete not working

## Demo
demo-games-panel

Related to https://trello.com/c/HSUZcSyK/37-3-as-norbert-i-want-to-be-able-to-see-all-games-and-be-able-to-delete-some

### DIFF
--- a/admin-app/.gitignore
+++ b/admin-app/.gitignore
@@ -1,24 +1,3 @@
-# Logs
-logs
-*.log
-npm-debug.log*
-yarn-debug.log*
-yarn-error.log*
-pnpm-debug.log*
-lerna-debug.log*
-
 node_modules
 dist
 dist-ssr
-*.local
-
-# Editor directories and files
-.vscode/*
-!.vscode/extensions.json
-.idea
-.DS_Store
-*.suo
-*.ntvs*
-*.njsproj
-*.sln
-*.sw?
--- a/admin-app/package.json
+++ b/admin-app/package.json
@@ -5,7 +5,9 @@
   "scripts": {
     "dev": "vite --port 5000 --host",
     "build": "tsc && vite build",
-    "preview": "vite preview"
+    "preview": "vite preview",
+    "test": "jest",
+    "test:watch": "jest --watch"
   },
   "dependencies": {
     "prop-types": "^15.8.1",
@@ -15,9 +17,13 @@
     "react-dom": "^17.0.2"
   },
   "devDependencies": {
+    "@testing-library/react": "^12.1.3",
+    "@types/jest": "^27.4.1",
     "@types/react": "^17.0.33",
     "@types/react-dom": "^17.0.10",
     "@vitejs/plugin-react": "^1.0.7",
+    "jest": "^27.5.1",
+    "ts-jest": "^27.1.3",
     "typescript": "^4.5.4",
     "vite": "^2.8.0"
   }
--- a/admin-app/package.json
+++ b/admin-app/package.json
@@ -11,7 +11,6 @@
   },
   "dependencies": {
     "prop-types": "^15.8.1",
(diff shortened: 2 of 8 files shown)


admin-app/.gitignore
```diff
@@ -0,0 +1,24 @@
...
+pnpm-debug.log*
+lerna-debug.log*
+
+node_modules
+dist
+dist-ssr
+*.local
+
+# Editor directories and files
+.vscode/*
```
fzaninotto: platform-specific stuff, these should not be in the project code

admin-app/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "private": true,
+  "version": "0.0.0",
+  "scripts": {
+    "dev": "vite --port 5000 --host",
+    "build": "tsc && vite build",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "prop-types": "^15.8.1",
+    "ra-data-fakerest": "^3.19.10",
```
fzaninotto: use the `^4.0.0-beta.1.0` version, or you'll have compatibility problems

admin-app/src/App.css
```diff
@@ -0,0 +1,42 @@
+.App {
```
fzaninotto: the way to style react-admin apps is via the sx prop, not a global CSS.

https://marmelab.com/blog/2022/02/27/mui-v5-rocks.html

fzaninotto (changes requested):

admin-app/src/App.css
atilbian: Ok, it's done in another branch, i will rebase on it.

admin-app/src/components/GameList.tsx
```diff
@@ -15,3 +22,14 @@ export const GameList = (props: GameListProps) => (
...
   </List>
 );
+
+const gameFilters = [
+  <SelectInput
+    source="status"
+    choices={[
+      { id: 'INITIALIZED', name: 'Initialized' },
+      { id: 'RUNNING', name: 'Running' },
+      { id: 'ENDED', name: 'Ended' },
```
slax57, line 32: Just for my information, did you have to add these values cause RA has no way of guessing them, or because the uppercase labels bothered you?

admin-app/src/data-provider.ts
```diff
@@ -16,44 +17,63 @@ const httpClient = (url: string, options: Options = {}) => {
...
+    const query = {
+      skip: (page - 1) * perPage,
+      take: perPage,
+      sort: [{ column: field, order: order }],
+      filter: Object.keys(params.filter).map((key) => ({
+        column: key,
+        value: params.filter[key],
+      })),
+    };
+    const url = `${adminApiUrl}/${resource}?s=${JSON.stringify(query)}`;
```
slax57: why not use `stringify` here as well? (for the 's' parameter)

admin-app/src/components/GameList.tsx
atilbian: RA can scrap the values inside the displayed lists of items but not guess other values. So I prefer be explicit define all the accepted values manually.  Ideally we would export the type from the nestjs entity(or extract it into a types file) but that's for another day !

admin-app/src/data-provider.ts
atilbian: Yep good idea !

PR #66, opened 2022-03-01
Add admin api find routes for games

slax57:
## Content

- [x] The computed **status** (INITIALIZED, RUNNING or ENDED) is now stored in db, in order to be able to filter on this column
- [x] Create a service with basic `findOne()` and `findMany()` implementations
- [x] Create a controller to expose these operations to REST API routes

Relates to: https://trello.com/c/HSUZcSyK/37-3-as-norbert-i-want-to-be-able-to-see-all-games-and-be-able-to-delete-some

## Demo
hex-api-admin-games

NB: Demo contains features that will only be available in the next PR https://github.com/marmelab/Hex/pull/67


web-app/src/games/api.admin.games.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+    const game = await this.gamesService.findOne(id);
+    if (game === undefined) throw new NotFoundException();
+    return game;
+  }
+
+  @Get('')
+  async findMany(
+    @Query('s', GamesSearchParamsPipe) searchParams?: GamesSearchParams,
+  ): Promise<Game[]> {
+    return await this.gamesService.findMany(searchParams);
```
fzaninotto: no need to await: you can return the Promise

fzaninotto (approved):

djhi (approved):

web-app/src/games/games.service.ts
```diff
@@ -90,20 +93,26 @@ export class GamesService {
...
-        ? 'RUNNING'
-        : 'INITIALIZED',
     };
     return gameAndStatus;
   }
+
+  computeStatusFromGame(game: Game): 'INITIALIZED' | 'RUNNING' | 'ENDED' {
+    return game.state.winner
+      ? 'ENDED'
+      : !!(game.player1 && game.player2)
```
atilbian: We can create a func to extract this redundant test

web-app/src/games/games.service.ts
```diff
@@ -13,11 +13,10 @@ import {
 import { Coordinates, deepCloneObject } from '../common/utils';
 import { UsersService } from '../users/users.service';
 
-export interface GameAndStatus {
+export interface GameAndDisplayStatus {
   game: Game;
   readyToPlay: boolean;
```
atilbian, line 18: This "readyToPlay" property is now redundant with the status defined inside the Game entity !

slax57: Indeed we should remove this attr and keep only `Game.status`. This would have impacts on both the web-app and the mobile-app, so I created this issue instead: https://github.com/marmelab/Hex/issues/68

PR #67, opened 2022-03-02
Complete admin api find routes for games with search criteria and delete routes

slax57:
## Content

- [x] Complete the `findMany()` implementation to handle search params
- [x] Complete the service with `deleteOne()` and `deleteMany()` implementations
- [x] Complete the controller
- [x] Cover the `convertToFindOptionsOrder()` and `convertToFindOptionsWhere()` functions with tests
- [x] Fix bug with `deleteMany()`, causing it to actually delete all games and not only the filtered ones
- [x] Add array support for filters (for bulk delete)

Relates to: https://trello.com/c/HSUZcSyK/37-3-as-norbert-i-want-to-be-able-to-see-all-games-and-be-able-to-delete-some

**Notes:** Any comments on the API contract (regarding the search params especially) are appreciated 🙂 

## Demo
hex-api-admin-games

### DIFF
--- a/web-app/src/games/api.admin.games.controller.ts
+++ b/web-app/src/games/api.admin.games.controller.ts
@@ -4,6 +4,8 @@ import {
   Param,
   UseGuards,
   NotFoundException,
+  Query,
+  Delete,
 } from '@nestjs/common';
 import { Game } from './game.entity';
 import { AuthGuard } from '@nestjs/passport';
--- a/web-app/src/games/api.admin.games.controller.ts
+++ b/web-app/src/games/api.admin.games.controller.ts
@@ -6,7 +6,15 @@ import {
   NotFoundException,
 } from '@nestjs/common';
 import { Game } from './game.entity';
-import { ApiAdminGamesService } from './api.admin.games.service';
+import { AuthGuard } from '@nestjs/passport';
+import {
+  ApiAdminGamesService,
+  GamesSearchParams,
+} from './api.admin.games.service';
+import {
+  GamesSearchParamsPipe,
+  JsonNumberArrayPipe,
+} from './api.admin.games.pipes';
 
 @Controller('api/admin/games')
 export class ApiAdminGamesController {
@@ -20,7 +28,19 @@ export class ApiAdminGamesController {
   }
 
   @Get('')
-  async findMany(): Promise<Game[]> {
-    return this.gamesService.findMany();
+  async findMany(
+    @Query('s', GamesSearchParamsPipe) searchParams?: GamesSearchParams,
+  ): Promise<Game[]> {
+    return this.gamesService.findMany(searchParams);
+  }
+
+  @Delete(':id')
+  deleteOne(@Param('id') id: number) {
+    this.gamesService.deleteOne(id);
+  }
+
+  @Delete('')
+  deleteMany(@Query('ids', JsonNumberArrayPipe) ids: number[]) {
+    this.gamesService.deleteMany(ids);
   }
 }
--- a/web-app/src/games/api.admin.games.controller.ts
+++ b/web-app/src/games/api.admin.games.controller.ts
@@ -13,7 +13,7 @@ import {
   ApiAdminGamesService,
   GamesSearchParams,
 } from './api.admin.games.service';
-import { JsonPipe } from './api.admin.games.pipes';
+import { SimpleJsonParsePipe } from './api.admin.games.pipes';
 
 @Controller('api/admin/games')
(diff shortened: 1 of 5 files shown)


web-app/src/games/api.admin.games.service.ts
```diff
@@ -14,7 +40,74 @@ export class ApiAdminGamesService {
...
+  deleteMany(ids: number[]) {
+    if (!ids) {
+      throw Error('A list of ids must be provided to delete games');
+    }
+    return this.gamesRepository.delete(ids);
+  }
+
+  convertToFindOptionsOrder(sort: GamesSortColumn[]): GameOrderType {
+    const order: GameOrderType = {};
+    sort.forEach((s) => {
```
atilbian: We can use plural here for the sort object, and give a better name for the var inside the loop !

web-app/src/games/api.admin.games.service.ts
```diff
@@ -14,7 +40,74 @@ export class ApiAdminGamesService {
...
+          order.player2 = s.order;
+          break;
+      }
+    });
+    return order;
+  }
+
+  convertToFindOptionsWhere(filter: GamesFilter[]): GameWhereType {
+    const where: GameWhereType = {};
+    filter.forEach((f) => {
```
atilbian: Same here for the plural + name of the loop var

web-app/src/games/api.admin.games.service.ts
```diff
@@ -14,7 +40,74 @@ export class ApiAdminGamesService {
...
+    }
+    return this.gamesRepository.find(findOptions);
+  }
+
+  deleteOne(id: number) {
+    return this.gamesRepository.delete(id);
+  }
+
+  deleteMany(ids: number[]) {
+    if (!ids) {
```
atilbian: Test length instead !

web-app/src/games/api.admin.games.controller.ts
```diff
@@ -21,8 +27,21 @@ export class ApiAdminGamesController {
...
+    return this.gamesService.findMany(searchParams);
+  }
+
+  @Delete(':id')
+  deleteOne(@Param('id') id: number) {
+    this.gamesService.deleteOne(id);
+  }
+
+  @Delete()
+  deleteMany(@Query('ids', new JsonPipe<number[]>()) ids: number[]) {
```
atilbian: There is a builtin ParseArrayPipe validators so we can use it instead to transform+validate query data

web-app/src/games/api.admin.games.pipes.ts
```diff
@@ -0,0 +1,12 @@
+import { PipeTransform, Injectable, ArgumentMetadata } from '@nestjs/common';
+
+@Injectable()
+export class JsonPipe<T> implements PipeTransform<any, T> {
```
atilbian: Do you agree to make it more explicit that this pipe is for data transformation only and not validation ?

web-app/src/games/api.admin.games.controller.ts
```diff
@@ -21,8 +27,21 @@ export class ApiAdminGamesController {
...
 
-  @Get('')
-  async findMany(): Promise<Game[]> {
-    return this.gamesService.findMany();
+  @Get()
+  async findMany(
+    @Query('s', new JsonPipe<GamesSearchParams>())
+    searchParams?: GamesSearchParams,
+  ): Promise<Game[]> {
+    return this.gamesService.findMany(searchParams);
```
atilbian, line 32: We need to return the total nb of entries in db to make the pagination work in the react admin ui !

web-app/src/games/api.admin.games.controller.ts
slax57: I tried to use it but couldn't get past through this error 🙁 :
```
[Nest] 104  - 03/03/2022, 10:03:16 AM   ERROR [PackageLoader] The "class-validator" package is missing. Please, make sure to install this library ($ npm install class-validator) to take advantage of ValidationPipe.
```

So I'll keep my pipe for now... 😕

web-app/src/games/api.admin.games.pipes.ts
slax57: Renamed it to `SimpleJsonParsePipe`, do you think it is enough to make this clear?

web-app/src/games/api.admin.games.controller.ts
slax57: Indeed!

This method now returns an object of the form: 

```
{ 
  data: Game[]; 
  total: number 
}
```

Issue #54, opened 2022-02-28
Fix authentication with wrong password


atilbian:
It's currently possible to authenticate without a correct password

slax57:
Fixed with https://github.com/marmelab/Hex/pull/73 🙂

PR #70, opened 2022-03-02
Protect admin routes to be used by the admin user only

slax57:
## Content

- [x] Fix login password check (https://github.com/marmelab/Hex/issues/54)
- [x] Add 'admin' role on admin api routes
- [x] Implement a RoleGuard
- [x] Store boolean to know if user is admin in JWT payload

Relates to: https://trello.com/c/XFCKJmht/39-05-as-norbert-i-want-to-be-the-only-person-able-to-access-the-administration-panel

### DIFF
--- a/web-app/src/auth/auth.service.ts
+++ b/web-app/src/auth/auth.service.ts
@@ -5,13 +5,13 @@ import { UsersService } from 'src/users/users.service';
 export interface UserDataInjectedInRequestAfterAuth {
   id: number;
   username: string;
-  admin?: boolean;
+  admin: boolean;
 }
 
 export interface JWTPayload {
   username: string;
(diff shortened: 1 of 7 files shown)


web-app/src/auth/auth.service.ts
```diff
@@ -5,11 +5,13 @@ import { UsersService } from 'src/users/users.service';
 export interface UserDataInjectedInRequestAfterAuth {
   id: number;
   username: string;
+  admin?: boolean;
 }
 
 export interface JWTPayload {
   username: string;
   sub: number;
+  admin?: boolean;
```
atilbian: Seems to me that this new field isn't nullable !

PR #71, opened 2022-03-02
Display a simple game list panel

atilbian:
- [x] Create a GameList component
- [x] Create a basic data provider with a getList function to fetch from our api
- [x] Store our api url into a env var

## Demo
demo-simple-games-panel

### DIFF
--- a/.github/workflows/deploy-admin-app-s3.yml
+++ b/.github/workflows/deploy-admin-app-s3.yml
@@ -24,7 +24,7 @@ jobs:
         run: yarn build
       - name: Create env file
         working-directory: ./admin-app
-        run: echo "$VITE_HEX_ADMIN_API_URL" > .env
+        run: echo "$ADMIN_ENV_VARS" > .env
       - name: Deploy to S3
         uses: jakejarvis/s3-sync-action@master
         with:
--- a/.github/workflows/deploy-admin-app-s3.yml
+++ b/.github/workflows/deploy-admin-app-s3.yml
@@ -24,7 +24,7 @@ jobs:
         run: yarn build
       - name: Create env file
         working-directory: ./admin-app
-        echo "$VITE_HEX_ADMIN_API_URL" > .env
+        run: echo "$VITE_HEX_ADMIN_API_URL" > .env
       - name: Deploy to S3
         uses: jakejarvis/s3-sync-action@master
         with:
--- a/.github/workflows/deploy-admin-app-s3.yml
+++ b/.github/workflows/deploy-admin-app-s3.yml
@@ -22,6 +22,9 @@ jobs:
       - name: Build
         working-directory: ./admin-app
         run: yarn build
+      - name: Create env file
+        working-directory: ./admin-app
+        echo "$VITE_HEX_ADMIN_API_URL" > .env
       - name: Deploy to S3
         uses: jakejarvis/s3-sync-action@master
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy-admin-app-s3.yml
```diff
@@ -22,6 +22,9 @@ jobs:
       - name: Build
         working-directory: ./admin-app
         run: yarn build
+      - name: Create env file
+        working-directory: ./admin-app
+        run: echo "$VITE_HEX_ADMIN_API_URL" > .env
```
slax57: You're mixing up the definition of **one** env var with the definition of a whole `.env` file.

I suggest stg like `echo "$ADMIN_ENV_VARS" > .env`
with `ADMIN_ENV_VARS` containing something like:

```
VITE_HEX_ADMIN_API_URL=https://d1iqp7uyjrmy2a.cloudfront.net/api/admin
```

admin-app/src/components/GameList.tsx
```diff
@@ -0,0 +1,16 @@
...
+interface GameListProps {}
+
+export const GameList = (props: GameListProps) => (
+  <List {...props}>
+    <Datagrid rowClick="edit">
+      <TextField source="id" />
+      <DateField source="createdAt" />
+      <DateField source="updatedAt" />
+      <NumberField source="player1.id" />
+      <NumberField source="player2.id" />
```
slax57, line 13: Can you confirm you did plan to add the 'status' column later?

admin-app/src/data-provider.ts
```diff
@@ -0,0 +1,47 @@
...
+    const url = `${apiUrl}/${resource}/`;
+
+    return httpClient(url).then(({ json }) => ({
+      data: json,
+      total: json.length,
+    }));
+  },
+
+  getOne: (resource: any) => {
+    throw Error();
```
slax57:
```suggestion
    throw Error('This function is not yet implemented');
```

And I suggest making this replacement for all unimplemented methods of this file of course 😉

.github/workflows/deploy-admin-app-s3.yml
atilbian: Oupsy doupsy !

admin-app/src/components/GameList.tsx
atilbian: Yep, but i can add it now too !

slax57 (approved):

admin-app/src/components/GameList.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+import { List, Datagrid, TextField, DateField, NumberField } from 'react-admin';
+
+interface GameListProps {}
+
+export const GameList = (props: GameListProps) => (
+  <List {...props}>
```
djhi, line 7: You don't need to pass props anymore in v4

admin-app/src/data-provider.ts
```diff
@@ -0,0 +1,47 @@
...
+const apiUrl = import.meta.env.VITE_HEX_ADMIN_API_URL;
+const httpClient = fetchUtils.fetchJson;
+
+export default {
+  getList: (resource: string) => {
+    const url = `${apiUrl}/${resource}/`;
+
+    return httpClient(url).then(({ json }) => ({
+      data: json,
+      total: json.length,
```
djhi, line 12: This is probably wrong but I suppose it's temporary

PR #76, opened 2022-03-03
Add getAll admin endpoint for users entities

atilbian:
- [x] Add endpoint with filtering, ordering and pagination capabilities

## Demo
demo-getall-users

Related to https://trello.com/c/IXbjh3Sd/40-1-as-norbert-i-want-to-be-able-to-see-everyone-using-my-app


web-app/src/users/api.admin.users.pipes.ts
```diff
@@ -0,0 +1,12 @@
+import { PipeTransform, Injectable, ArgumentMetadata } from '@nestjs/common';
+
+@Injectable()
+export class SimpleJsonParsePipe<T> implements PipeTransform<any, T> {
```
slax57: Why duplicate this pipe? We should rather move the existing pipe to a common folder, like a 'utils' folder

web-app/src/users/api.admin.users.service.ts
```diff
@@ -0,0 +1,97 @@
+import { Injectable } from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
+import { FindConditions, FindManyOptions, Repository } from 'typeorm';
+import { EntityFieldsNames } from 'typeorm/common/EntityFieldsNames';
+import { User } from './user.entity';
+
+export type SortOrder = 'ASC' | 'DESC';
```
slax57: We could extract this type as well

PR #83, opened 2022-03-07
API: Add function to determine if I can win the game in up to 1 move

slax57:
- [x] Add a common function that determines, given a game state and a player, if the game is won or if someone can win it with one move (also returns the move if that is the case)
- [x] Covered function with unit tests
- _Probably in next PR_: Add API route to call this function and see it in action on a real game state

Related to: https://trello.com/c/wTvhubhQ/29-05-as-norbert-given-a-game-state-i-want-to-know-if-i-have-a-chance-of-winning-or-not-on-the-next-turn

### DIFF
--- a/web-app/src/common/gameState.spec.ts
+++ b/web-app/src/common/gameState.spec.ts
@@ -4,6 +4,8 @@ import {
   playerHasWon,
   initNewGameState,
   updateGameState,
+  getNextMoveHint,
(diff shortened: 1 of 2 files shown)


guilbill (approved):
Despite the `closenessToGameEnd` variable, which is not enough self explanable in my opinion, that's all good.
Love the readability of the test ;)

PR #84, opened 2022-03-07
Calculate winning path cost for a given board state

atilbian:
- [x] Calculate the cost of the shortest path
- [x] Add edges for empty cells too
- [x] Use unidirictionnal edges instead of bidirectionnal edges and add a different cost for empty and cells of the targeted color
- [x] Test the calculation of the winning path cost
- [x] Wait for https://github.com/marmelab/Hex/pull/84 to be merged

Related to https://trello.com/c/0QadYH4a/30-1-as-norbert-given-a-game-state-i-want-a-better-advice-on-the-next-move

### DIFF
--- a/web-app/src/common/graphWinPrediction.ts
+++ b/web-app/src/common/graphWinPrediction.ts
@@ -75,13 +75,19 @@ function createPlayableCellEdges(
           { y: y - 1, x: x + 1 }, { y: y + 1, x: x - 1 },
         ];
         directNeighborCells.forEach(neighbor => {
-          if (!doesCellExist(gameState, neighbor) || doesCellExistAndHaveStone(gameState, neighbor, stoneColor === "black" ? "white" : "black")) return;
-          const cost = doesCellExistAndHaveStone(gameState, neighbor, stoneColor) ? LOW_EDGE_COST : HIGH_EDGE_COST;
-          addEdgeWithCost(hexBoardWinPredictionGraph, `${currentCell.y}-${currentCell.x}`, `${neighbor.y}-${neighbor.x}`, cost);
+          if (doesCellCouldBeOnVictoryPath(gameState, stoneColor, neighbor)) {
+            const cost = doesCellExistAndHaveStone(gameState, neighbor, stoneColor) ? LOW_EDGE_COST : HIGH_EDGE_COST;
+            addEdgeWithCost(hexBoardWinPredictionGraph, `${currentCell.y}-${currentCell.x}`, `${neighbor.y}-${neighbor.x}`, cost);
+          }
         })
       }
     })
   });
+}
+
+function doesCellCouldBeOnVictoryPath(gameState: GameState, stoneColor: StoneColor, coordinates: Coordinates) {
+  const opponentColor = stoneColor === "black" ? "white" : "black";
+  return (!doesCellExist(gameState, coordinates) || doesCellExistAndHaveStone(gameState, coordinates, opponentColor)) ? false : true;
 }
 
 /**
--- a/web-app/src/common/graphWinPrediction.ts
+++ b/web-app/src/common/graphWinPrediction.ts
@@ -1,0 +1,125 @@
+import * as jkstra from 'jkstra';
+import {
+  doesCellExist, doesCellExistAndHaveStone, GameState, StoneColor,
+  BLACK_NODE_START, BLACK_NODE_END, WHITE_NODE_START, WHITE_NODE_END
+} from './gameState';
+import { Coordinates } from './utils';
+
+export const UNDEFINED_EDGE_COST = -1;
+export const LOW_EDGE_COST = 1;
+export const HIGH_EDGE_COST = 100;
+
+export interface HexBoardWinPredictionGraph {
+  winPrediction: jkstra.Graph;
+  vertexMap: Map<string, jkstra.Vertex>;
+}
+
+/**
+ * Create a graph to represent the winnable paths of a given board.
+ * It doesn't add vertex/edges for the cells of the opposite color because it's not needed for this calculation.
+ */
+export function createWinPredictionGraph(
+  gameState: GameState,
+  stoneColor: StoneColor,
+): HexBoardWinPredictionGraph {
+  const hexBoardWinPredictionGraph: HexBoardWinPredictionGraph = {
+    winPrediction: new jkstra.Graph(),
+    vertexMap: new Map<string, jkstra.Vertex>(),
+  };
+  createVertices(gameState, hexBoardWinPredictionGraph, stoneColor);
+  createStartEndEdges(gameState, hexBoardWinPredictionGraph, stoneColor);
+  createPlayableCellEdges(gameState, hexBoardWinPredictionGraph, stoneColor);
+  return hexBoardWinPredictionGraph;
+}
+
+function createVertices(
+  gameState: GameState,
+  hexBoardWinPredictionGraph: HexBoardWinPredictionGraph,
+  stoneColor: StoneColor
+) {
+  if (stoneColor === "black") {
+    addVertex(hexBoardWinPredictionGraph, BLACK_NODE_START);
+    addVertex(hexBoardWinPredictionGraph, BLACK_NODE_END);
+  }
+  else {
+    addVertex(hexBoardWinPredictionGraph, WHITE_NODE_START);
+    addVertex(hexBoardWinPredictionGraph, WHITE_NODE_END);
+  }
+
+  // Add node for each cell on the board that's are not of the opposite color
+  gameState.board.forEach((row, y) => {
+    row.forEach((_cell, x) => {
+      if (_cell.value === stoneColor || _cell.value === "empty") {
+        addVertex(hexBoardWinPredictionGraph, `${y}-${x}`);
+      }
+    });
+  });
+}
+
+function addVertex(hexBoardWinPredictionGraph: HexBoardWinPredictionGraph, id: string) {
+  hexBoardWinPredictionGraph.vertexMap.set(id, hexBoardWinPredictionGraph.winPrediction.addVertex(id));
+}
+
+function createPlayableCellEdges(
+  gameState: GameState,
+  hexBoardWinPredictionGraph: HexBoardWinPredictionGraph,
+  stoneColor: StoneColor,
+) {
+  gameState.board.forEach((row, y) => {
+    row.forEach((cell, x) => {
+      if (cell.value == stoneColor || cell.value == "empty") {
+        const currentCell: Coordinates = { y: y, x: x };
+        const directNeighborCells = [
+          { y: y, x: x - 1 }, { y: y, x: x + 1 },
+          { y: y - 1, x: x }, { y: y + 1, x: x },
+          { y: y - 1, x: x - 1 }, { y: y + 1, x: x + 1 },
+        ];
+        directNeighborCells.forEach(neighbor => {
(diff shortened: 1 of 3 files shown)


web-app/src/common/graphWinPrediction.ts
```diff
@@ -0,0 +1,131 @@
...
+import { Coordinates } from './utils';
+
+export const UNDEFINED_EDGE_COST = -1;
+export const LOW_EDGE_COST = 1;
+export const HIGH_EDGE_COST = 100;
+
+export interface HexBoardWinPredictionGraph {
+  winPrediction: jkstra.Graph;
+  vertexMap: Map<string, jkstra.Vertex>;
+}
```
slax57, line 15: I don't understand why you need both this interface and this one:

```tx
export interface HexBoardWinDetectionGraph {
  winDetectionGraph: jkstra.Graph;
  vertexMap: Map<string, jkstra.Vertex>;
}
```

Can't you factorize?

Moreover, I found the code to be clearer when the first attribute was called `graph`. Here, in your code, you have things like `winPrediction.addVertex(id)` which I find less readable than  `graph.addVertex(id)` .

web-app/src/common/graphWinPrediction.ts
```diff
@@ -0,0 +1,131 @@
...
+) {
+  gameState.board.forEach((row, y) => {
+    row.forEach((cell, x) => {
+      if (cell.value == stoneColor || cell.value == "empty") {
+        const currentCell: Coordinates = { y: y, x: x };
+        const directNeighborCells = [
+          { y: y, x: x - 1 }, { y: y, x: x + 1 },
+          { y: y - 1, x: x }, { y: y + 1, x: x },
+          { y: y - 1, x: x - 1 }, { y: y + 1, x: x + 1 },
+        ];
```
slax57: I'm a bit surprised by this. I expected something like this:

```suggestion
        const directNeighborCells = [
          { y: y, x: x - 1 }, { y: y, x: x + 1 },
          { y: y - 1, x: x }, { y: y + 1, x: x },
          { y: y - 1, x: x + 1 }, { y: y + 1, x: x - 1 },
        ];
```

I'm gonna run some tests, maybe I'm mistaken...

web-app/src/common/graphWinPrediction.ts
```diff
@@ -0,0 +1,131 @@
...
+        directNeighborCells.forEach(neighbor => {
+          if (!doesCellExist(gameState, neighbor) || doesCellExistAndHaveStone(gameState, neighbor, stoneColor === "black" ? "white" : "black")) return;
```
slax57: This line is too complicated, it could be extracted in a method, allowing you to pick a name explaining clearly what it does.

Also, for something like `stoneColor === "black" ? "white" : "black"`, the same principle applies. You can for instance define a constant to hold its value, and clarify what it does, like this:

```ts
const opponentColor = stoneColor === "black" ? "white" : "black";
```

web-app/src/common/graphWinPrediction.ts
```diff
@@ -0,0 +1,131 @@
...
+      const currentFirstRowCellValue = gameState.board[0][i].value;
+      if (currentFirstRowCellValue !== "black") {
+        addEdgeWithCost(hexBoardWinPredictionGraph, WHITE_NODE_START, `0-${i}`, currentFirstRowCellValue === "white" ? LOW_EDGE_COST : HIGH_EDGE_COST);
+      }
+      const currentLastRowCellValue = gameState.board[gameState.board.length - 1][i].value;
+      if (currentLastRowCellValue !== "black") {
+        addEdgeWithCost(hexBoardWinPredictionGraph, `${gameState.board.length - 1}-${i}`, WHITE_NODE_END, UNDEFINED_EDGE_COST);
+      }
+    }
+  });
```
slax57, line 123: I get what you did there, but it is a bit hard to read and understand imo.

One way you could have improved this (but I'm not sure it's worth the refactoring though), is to consider this:
- For now, your function `addEdgeWithCost()` simply does what it's told, ie. it creates an edge with a cost.
- But I believe it could do more than that, since the behaviour is always the same:
  - Create the edge only if the starting node exists and is not white
  - Create the edge only if the end node exists and is not white
  - Compute the cost depending on the end node: if empty, HIGH COST, otherwise, LOW COST
- with that in mind, you could delegate all these checks to the child function, and then your `createStartEndEdges()` function would be much simpler

web-app/src/common/pathfinding.ts
```diff
@@ -28,3 +28,24 @@ export function getWinningPathIfExist(
...
+      edgeCost: function (edge) {
+        return edge.data;
+      }
+    });
+  if (!shortestPath) return -1;
+  const cost = Object.keys(shortestPath).reduce(function (previous, key) {
+    return shortestPath[key].data === UNDEFINED_EDGE_COST ? previous : previous + shortestPath[key].data
+  }, 0);
+  return cost;
+}
```
slax57: Can't this function already return the number of moves needed to win, instead of this 'cost' which, to me, is more of a computation intermediate than a real result?

web-app/src/common/pathfinding.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+    const input = parseGameStateFromMultilineString(`
+  W ⬡ ⬡
+   ⬡ W ⬡
+    ⬡ ⬡ W
+  `);
+
+    const cost = getWinnablePathCost(input, "white");
+
+    expect(cost).toStrictEqual(3);
+  });
```
slax57: I disagree with this test. To me, these white stones are **not** connected.

slax57 (changes requested):

web-app/src/common/pathfinding.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+   ⬢ ⬢ ⬢
+    ⬢ ⬢ ⬢
+  `);
+
+    const cost = getWinnablePathCost(input, "black");
+
+    expect(cost).toStrictEqual(3);
+  });
+
+  it('Should return a valid winning path cost', () => {
```
slax57: One more thing, all your tests have the same 'it' description. I believe this is a bad practice.

slax57 (changes requested):

web-app/src/common/graphWinPrediction.ts
slax57: As I feared, I believe you made a mistake here 🙁 .

This test should pass, but it doesn't:

```ts
  it('Should return a valid winning path cost hum...', () => {
    const input = parseGameStateFromMultilineString(`
  ⬡ ⬡ W
   ⬡ W ⬡
    W ⬡ ⬡
  `);

    const cost = getWinnablePathCost(input, 'white');

    expect(cost).toStrictEqual(3);
  });
```

It returns a cost of 201 instead of 3

web-app/src/common/graphWinPrediction.ts
atilbian: I voluntarily chose to split it in 2 differents interfaces to be explicit about the diff between the 2 kinds of graph and avoid using them in the wrong context.
The use of a specific name for the graph property(like winPrediction instead of graph) is to avoid the [duck-typing](https://www.javatpoint.com/typescript-duck-typing) TypeScript feature ! But we could use another name, as long as they are differents in the 2 interfaces

web-app/src/common/graphWinPrediction.ts
atilbian: Good catch !

web-app/src/common/pathfinding.spec.ts
atilbian: Yep i agree with that, but have you got a suggestion to improve the descr without repeating the expected value ?

web-app/src/common/pathfinding.spec.ts
atilbian: Same mistake as here : https://github.com/marmelab/Hex/pull/84#discussion_r821708086

web-app/src/common/graphWinPrediction.ts
atilbian: Yes, I agree with you, that part is not very readable and probably needs to be reworked. But I'm not sure I understand your solution because addEdgeWithCost is also used to draw edges between common cells so I think it's good to let it do the minimal work without adding more complexity to it.

web-app/src/common/pathfinding.ts
atilbian: Yes !

web-app/src/common/graphWinPrediction.ts
```diff
@@ -0,0 +1,137 @@
...
+    row.forEach((cell, x) => {
+      if (cell.value == stoneColor || cell.value == "empty") {
+        const currentCell: Coordinates = { y: y, x: x };
+        const directNeighborCells = [
+          { y: y, x: x - 1 }, { y: y, x: x + 1 },
+          { y: y - 1, x: x }, { y: y + 1, x: x },
+          { y: y - 1, x: x + 1 }, { y: y + 1, x: x - 1 },
+        ];
+        directNeighborCells.forEach(neighbor => {
+          if (doesCellCouldBeOnVictoryPath(gameState, stoneColor, neighbor)) {
```
fzaninotto: this is not English

web-app/src/common/graphWinPrediction.ts
slax57: I still don't agree fully, since I don't really see any place where we could confuse `HexBoardWinDetectionGraph` with `HexBoardWinPredictionGraph` (in the end, in `pathfinding.ts`, we just export 2 functions with different signatures and names, so there is no risk of confusion, and the graphs are only used internally for calculations).

But on the other hand, I agree that this is not really important. At least, for now. These two interfaces can co-exist, and I don't think it will cause huge maintenance problems later either. Let's leave it that way.

web-app/src/common/graphWinPrediction.ts
slax57: Unless I'm missing something, even in the case of 'common cells', the function as I describe it would still work, wouldn't it?

Anyways, as I said,  I'm not sure it's worth the refactoring for so little. Let's keep it that way for now.

web-app/src/common/pathfinding.spec.ts
slax57: I see no other way 🤷

web-app/src/common/graphWinPrediction.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+  else {
+    addVertex(hexBoardWinPredictionGraph, WHITE_NODE_START);
+    addVertex(hexBoardWinPredictionGraph, WHITE_NODE_END);
+  }
+
+  // Add vertex for each cell on the board that's are not of the opposite color
+  gameState.board.forEach((row, y) => {
+    row.forEach((_cell, x) => {
+      if (_cell.value === stoneColor || _cell.value === "empty") {
```
slax57: `_cell` is a convention to mark unused variables. If you're actually gonna use it, then remove the underscore

web-app/src/common/pathfinding.ts
```diff
@@ -28,3 +28,35 @@ export function getWinningPathIfExist(
...
   return { hasWon: false, winningPath: null };
 }
+
+export function getNbMovesNeededToWin(
+  gameState: GameState,
+  stoneColor: StoneColor,
+): number {
+  const hexBoardWinPredictionGraph = createWinPredictionGraph(gameState, stoneColor);
+  const dijkstra = new jkstra.algos.Dijkstra(hexBoardWinPredictionGraph.winPrediction);
+  let shortestPath = dijkstra.shortestPath(
```
slax57:
```suggestion
  const shortestPath = dijkstra.shortestPath(
```

slax57 (approved):

PR #87, opened 2022-03-08
Get next play advice

atilbian:
- [x] Add a func to give a good play on a given board
- [x] Test it
- [x] Plug this func on the existing getNextMoveHint func
- [x] Wait for https://github.com/marmelab/Hex/pull/84 to be merged

## Demo
demo-play-suggestions

Related to https://trello.com/c/0QadYH4a/30-1-as-norbert-given-a-game-state-i-want-a-better-advice-on-the-next-move

### DIFF
--- a/web-app/src/common/prediction.spec.ts
+++ b/web-app/src/common/prediction.spec.ts
@@ -1,0 +1,79 @@
+import expect from 'expect';
+import { parseGameStateFromMultilineString } from './utils';
+import { getNextPlayAdvice } from './prediction';
+import exp from 'constants';
+
+describe('Get an advice for the next play', () => {
+  it('Get a good advice', () => {
+    const input = parseGameStateFromMultilineString(`
+  ⬡ ⬡ ⬡
+   ⬡ ⬡ ⬡
+    ⬢ ⬢ ⬡
+  `);
+
+    const nextPlayAdvice = getNextPlayAdvice(input, "black");
+
+    expect(nextPlayAdvice).toStrictEqual({ x: 2, y: 2 });
+  });
+
+  it('Get a good advice', () => {
+    const input = parseGameStateFromMultilineString(`
+  ⬡ ⬡ ⬡
+   ⬡ ⬡ ⬡
+    ⬢ ⬡ ⬢
+  `);
+
+    const nextPlayAdvice = getNextPlayAdvice(input, "black");
+
+    expect(nextPlayAdvice).toStrictEqual({ x: 1, y: 2 });
+  });
+
+
+  it('Get a good advice', () => {
+    const input = parseGameStateFromMultilineString(`
+  ⬡ ⬡ ⬡
+   ⬡ ⬡ ⬡
+    ⬡ ⬢ ⬢
+  `);
+
+    const nextPlayAdvice = getNextPlayAdvice(input, "black");
+
+    expect(nextPlayAdvice).toStrictEqual({ x: 0, y: 2 });
+  });
+
+  it('Get a good advice', () => {
+    const input = parseGameStateFromMultilineString(`
+  ⬡ ⬡ ⬡
+   W ⬡ ⬡
+    W ⬡ ⬡
+  `);
+
+    const nextPlayAdvice = getNextPlayAdvice(input, "white");
+
+    expect(nextPlayAdvice).toStrictEqual({ x: 0, y: 0 });
+  });
+
+  it('Get a good advice', () => {
+    const input = parseGameStateFromMultilineString(`
+  W ⬡ ⬡
+   ⬡ ⬡ ⬡
+    W ⬡ ⬡
+  `);
+
+    const nextPlayAdvice = getNextPlayAdvice(input, "white");
+
+
+  });
+
+  it('Get a good advice', () => {
+    const input = parseGameStateFromMultilineString(`
+  W ⬡ ⬡
+   W ⬡ ⬡
+    ⬡ ⬡ ⬡
+  `);
+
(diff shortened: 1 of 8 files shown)


web-app/src/common/prediction.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  it('Should get a good advice', () => {
+    const input = parseGameStateFromMultilineString(`
+  ⬡ ⬡ ⬡
+   ⬡ ⬡ ⬡
+    ⬢ ⬡ ⬢
+  `);
+
+    const nextPlayAdvice = getNextPlaySuggestion(input, "black");
+
+    expect(areCoordinatesEquals(nextPlayAdvice, { x: 1, y: 2 }) || areCoordinatesEquals(nextPlayAdvice, { x: 1, y: 1 })).toBeTruthy();
```
slax57: I disagree on this one. To me, playing at `{ x: 1, y: 1 }` doesn't make a connection, only `{ x: 1, y: 2 }` is valid

web-app/src/common/prediction.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  it('Should get a good advice', () => {
+    const input = parseGameStateFromMultilineString(`
+  ⬡ ⬡ ⬡
+   ⬡ ⬡ ⬡
+    ⬡ ⬢ ⬢
+  `);
+
+    const nextPlayAdvice = getNextPlaySuggestion(input, "black");
+
+    expect(areCoordinatesEquals(nextPlayAdvice, { x: 0, y: 2 }) || areCoordinatesEquals(nextPlayAdvice, { x: 0, y: 1 })).toBeTruthy();
```
slax57: Same here, only `{ x: 0, y: 2 }` is valid according to me

web-app/src/common/prediction.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  it('Should get a good advice', () => {
+    const input = parseGameStateFromMultilineString(`
+  W ⬡ ⬡
+   ⬡ ⬡ ⬡
+    W ⬡ ⬡
+  `);
+
+    const nextPlayAdvice = getNextPlaySuggestion(input, "white");
+
+    expect(areCoordinatesEquals(nextPlayAdvice, { x: 0, y: 1 }) || areCoordinatesEquals(nextPlayAdvice, { x: 1, y: 1 })).toBeTruthy();
```
slax57: Same here, only `{ x: 0, y: 1 }` is valid

web-app/src/common/prediction.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  it('Should get a good advice', () => {
+    const input = parseGameStateFromMultilineString(`
+  W ⬡ ⬡
+   W ⬡ ⬡
+    ⬡ ⬡ ⬡
+  `);
+
+    const nextPlayAdvice = getNextPlaySuggestion(input, "white");
+
+    expect(areCoordinatesEquals(nextPlayAdvice, { x: 0, y: 2 }) || areCoordinatesEquals(nextPlayAdvice, { x: 1, y: 2 })).toBeTruthy();
```
slax57: same

web-app/src/common/prediction.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  it('Should get a good advice', () => {
+    const input = parseGameStateFromMultilineString(`
+  W ⬡ ⬡
+   ⬡ W ⬡
+    ⬡ ⬡ ⬡
+  `);
+
+    const nextPlayAdvice = getNextPlaySuggestion(input, "white");
+
+    expect(areCoordinatesEquals(nextPlayAdvice, { x: 1, y: 2 }) || areCoordinatesEquals(nextPlayAdvice, { x: 2, y: 2 })).toBeTruthy();
```
slax57: Here it's different, I would expect `{ x: 0, y: 2 }` or `{ x: 1, y: 2 }` but not `{ x: 2, y: 2 }`

web-app/src/common/prediction.ts
```diff
@@ -0,0 +1,24 @@
...
+  if (allPotentialPlays.length === 0) throw new Error("There is no playable cell in the given board.")
+
+  const winCostForPotentialPlays = allPotentialPlays.map(playPosition => {
+    const potentialGameState = deepCloneObject(gameState) as GameState;
+    potentialGameState.board[playPosition.y][playPosition.x].value = stoneColor;
+    return { playPosition, cost: getNbMovesNeededToWin(potentialGameState, stoneColor) }
+  });
+
+  return winCostForPotentialPlays.reduce(function (prev, curr) {
+    return prev.cost < curr.cost ? prev : curr;
```
slax57: I was wondering why your algo always returned cells closer to the right or bottom edge, but now I have my answer 😁 (`<` instead of `<=`). I find it a bit counter-intuitive for the IA to suggest moves that way, but it's not wrong either 😁 . So I guess I'm okay with this 😂

slax57 (changes requested):

web-app/src/common/prediction.spec.ts
atilbian: I think `{ x: 1 y: 0 }` , `{ x: 2 y: 0}` and `{ x: 0 y: 1}` are valid too !

web-app/src/common/prediction.ts
atilbian: I switch to give the priority to upper left cells !

slax57 (approved):

PR #88, opened 2022-03-08
Create board binary representation

slax57:
## Content

- [x] Add function to convert an array representation of the board to a binary one
- [x] Add function to do the opposite conversion

Related to: https://trello.com/c/eOslB09A/77-1-as-norbert-i-want-a-more-intelligent-ai

### DIFF
--- a/web-app/src/common/utils.spec.ts
+++ b/web-app/src/common/utils.spec.ts
@@ -1,0 +1,73 @@
+import expect from 'expect';
+import {
+  arrayBoardToBinaryBoard,
+  parseGameStateFromMultilineString,
+} from './utils';
+
+describe('convert an array representation of the board to a binary one', () => {
+  it('should convert this 3x3 array board into its binary representation', () => {
+    // given
+    const input = parseGameStateFromMultilineString(`
+⬡ W ⬡
+ ⬢ W ⬡
+  ⬢ W ⬢
+        `).board;
+    // when
+    const ouput = arrayBoardToBinaryBoard(input);
+    // then
+    expect(ouput).toStrictEqual({
+      whiteBoard: 0b010010010,
+      blackBoard: 0b000100101,
+    });
+  });
+
+  it('should convert this 3x3 array board with only black stones into its binary representation', () => {
+    // given
+    const input = parseGameStateFromMultilineString(`
+⬢ ⬢ ⬢
+ ⬢ ⬢ ⬢
+  ⬢ ⬢ ⬢
+        `).board;
+    // when
+    const ouput = arrayBoardToBinaryBoard(input);
+    // then
+    expect(ouput).toStrictEqual({
+      whiteBoard: 0b000000000,
+      blackBoard: 0b111111111,
+    });
+  });
+
+  it('should convert this 3x3 array board with only white stones into its binary representation', () => {
+    // given
+    const input = parseGameStateFromMultilineString(`
+W W W
+ W W W
+  W W W
+        `).board;
+    // when
+    const ouput = arrayBoardToBinaryBoard(input);
+    // then
+    expect(ouput).toStrictEqual({
+      whiteBoard: 0b111111111,
+      blackBoard: 0b000000000,
+    });
+  });
+
+  it('should convert this 5x5 array board into its binary representation', () => {
+    // given
+    const input = parseGameStateFromMultilineString(`
+⬡ W ⬡ ⬢ ⬡
+ ⬢ W ⬡ ⬡ ⬡
+  ⬢ W ⬢ W W
+   ⬡ ⬡ ⬡ ⬡ W
+    ⬢ ⬡ ⬡ ⬡ ⬢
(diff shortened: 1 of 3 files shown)


web-app/src/common/utils.ts
```diff
@@ -48,3 +54,54 @@ export function parseGameStateFromMultilineString(
...
+  color: StoneColor,
+): number {
+  return board.reduce(
+    (prevRow, row, y) =>
+      prevRow +
+      row.reduce(
+        (prev, cell, x) =>
+          prev +
+          (cell.value === color
+            ? Math.pow(2, boardSize - 1 - (y * row.length + x))
```
fzaninotto: very inefficient. Can't you do that with binary operators?

web-app/src/common/utils.ts
```diff
@@ -48,3 +54,54 @@ export function parseGameStateFromMultilineString(
...
+}
+
+function arrayBoardToBinaryBoardByColor(
+  board: Array<Array<Cell>>,
+  boardSize: number,
+  color: StoneColor,
+): number {
+  return board.reduce(
+    (prevRow, row, y) =>
+      prevRow +
```
fzaninotto: can't you replace a `+` by a boolean `or`?

web-app/src/common/utils.ts
```diff
@@ -48,3 +54,54 @@ export function parseGameStateFromMultilineString(
   parsedGameState.winningPath = getWinnerDataIfExist.winningPath;
   return parsedGameState;
 }
+
+export function arrayBoardToBinaryBoard(
+  board: Array<Array<Cell>>,
+): BinaryBoard {
+  const boardSize = board.length * board[0].length;
```
fzaninotto: I don't understand why the boardsize is the total number of hex rather than the width. It forces you to do sqrt later...

web-app/src/common/utils.ts
```diff
@@ -48,3 +54,54 @@ export function parseGameStateFromMultilineString(
...
+
+export function binaryBoardToArrayBoard(
+  board: BinaryBoard,
+): Array<Array<Cell>> {
+  const rowLength = Math.sqrt(board.boardSize);
+  const result: Array<Array<Cell>> = [];
+  for (let y = 0; y < rowLength; y++) {
+    result[y] = [];
+    for (let x = 0; x < rowLength; x++) {
+      const filter: number = Math.pow(
```
fzaninotto: same, try to do the same with no math operators, otherwise it's super slow

fzaninotto (changes requested):
I'm not sure your optimization is an optimization. I advise you benchmark the 2 algos (e.g. by creating a random 19x19 board, and converting it 1000 times to binary)

web-app/src/common/utils.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+
+describe('convert an array representation of the board to a binary one', () => {
+  it('should convert this 3x3 array board into its binary representation', () => {
+    // given
+    const input = parseGameStateFromMultilineString(`
+⬡ W ⬡
+ ⬢ W ⬡
+  ⬢ W ⬢
+        `).board;
+    // when
```
atilbian: Didn't I win the discussion about these comments :p ?

web-app/src/common/utils.ts
```diff
@@ -48,3 +54,48 @@ export function parseGameStateFromMultilineString(
...
+  const boardSize = board.length;
+  return {
+    whiteBoard: arrayBoardToBinaryBoardByColor(board, boardSize, 'white'),
+    blackBoard: arrayBoardToBinaryBoardByColor(board, boardSize, 'black'),
+    boardSize,
+  };
+}
+
+function arrayBoardToBinaryBoardByColor(
+  board: Array<Array<Cell>>,
```
atilbian: I think we can use a type here !

web-app/src/common/utils.ts
```diff
@@ -1,9 +1,15 @@
...
+import { Cell, GameState, getWinner, StoneColor } from './gameState';
 export interface Coordinates {
   x: number;
   y: number;
 }
 
+export interface BinaryBoard {
+  whiteBoard: number;
+  blackBoard: number;
+  boardSize: number;
```
atilbian, line 10: I prefer recalculating it instead of having a redundant value ! I think it's not so much because it's done only one time per conversion.

atilbian: Ok as discussed in private this field is required so i close this comment.

web-app/src/common/utils.spec.ts
slax57: Sorry about that. Copy pasted `gameState.spec.ts` and forgot to remove them later 👼

atilbian:
> I'm not sure your optimization is an optimization. I advise you benchmark the 2 algos (e.g. by creating a random 19x19 board, and converting it 1000 times to binary)

PR #89, opened 2022-03-09
Improve next play advice based on the opponent score

atilbian:
- [x] Find a metric to take the opponent into account for the suggested move
- [x] Update the func to suggest a move to use this new metric
- [x] Add more relevant tests
- [x] Wait for https://github.com/marmelab/Hex/pull/87 to be merged
- [x] ~Think about what to do when the next move is a winning move for the opponent(merge the code from the hint logic into this suggestion func ?)~ => Moved into the future work about the AI

Related to https://trello.com/c/fYq01Sfi/33-05-as-norbert-i-want-to-be-advised-on-what-best-move-i-should-play-next-considering-patricks-options

## Demo
demo-play-suggestion-based-on-opponent-too

### DIFF
--- a/web-app/src/common/prediction.spec.ts
+++ b/web-app/src/common/prediction.spec.ts
@@ -115,7 +115,7 @@ describe('Get a suggestion for the next play', () => {
     expect(() => getNextPlaySuggestion(input, "white")).toThrowError();
   });
 
-  it('Should get a suggestion to play on x:3,y:1 based on the two players', () => {
+  it('Should get a suggestion to play on x:3,y:1', () => {
     const input = parseGameStateFromMultilineString(`
     ⬡ ⬡ ⬡ ⬡ ⬡
      ⬡ ⬡ ⬡ ⬡ ⬡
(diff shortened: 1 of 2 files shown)


web-app/src/common/prediction.spec.ts
```diff
@@ -114,4 +114,37 @@ describe('Get a suggestion for the next play', () => {
...
+
+    const nextPlayAdvice = getNextPlaySuggestion(input, "white");
+
+    expect(nextPlayAdvice).toStrictEqual({ x: 3, y: 1 });
+  });
+
+  it('Should get a suggestion to play on one from 3 possible coordinates', () => {
+    const input = parseGameStateFromMultilineString(`
+    ⬡ W ⬡ ⬡ ⬡
+     ⬡ ⬡ ⬡ E ⬢
```
slax57: Why use `E` instead of `⬡` 🤔

atilbian: Oooops

web-app/src/common/prediction.spec.ts
```diff
@@ -114,4 +114,37 @@ describe('Get a suggestion for the next play', () => {
...
+    ⬡ W ⬡ ⬡ ⬡
+     ⬡ ⬡ ⬡ ⬡ ⬢
+      W ⬡ ⬢ W ⬡
+       ⬡ ⬢ ⬡ W ⬡
+        ⬡ ⬡ ⬡ ⬡ ⬡
+    `);
+
+    const nextPlayAdvice = getNextPlaySuggestion(input, "black");
+
+    expect(areExpectedCoordinatesInList(nextPlayAdvice, [
```
guilbill, line 143: I didn't see it before, but no need to define a function for that :
```expect([
    { x: 3, y: 1 },
    { x: 0, y: 3 },
    { x: 0, y: 4 },
]).toContainEqual(nextPlayAdvice)```

Advantage : 
- Test report is clearer
- No utils method to maintain
- The method from jest expect is bullet proof (yours is not...)

In general, take some time to find if the assertion library can check what you want to, before adding a custom one.

File: web-app/src/common/prediction.ts [hunk 3]
```diff
@@ -2,23 +2,46 @@ import { GameState, StoneColor } from './gameState';
...
 import { getNbMovesNeededToWin } from './pathfinding';
 
+interface PlayPrediction {
+  coordinates: Coordinates;
+  playerRemainingMovesToWin: number;
+  opponentRemainingMovesToWin: number;
+  score: number;
+}
+
 export function getNextPlaySuggestion(gameState: GameState, stoneColor: StoneColor): Coordinates {
```
guilbill, line 12: Just a formating matter, but I find it a little too dense, if you add some blank line (in a logical way, like blank lines around the conditional throw) it can ease the reading:
```
  const playableCells = getPlayableCells(gameState);
  
  if (playableCells.length === 0) throw new Error("There is no playable cell in the given board.")

  const playPredictions = playableCells.map(coordinates =>
    getPlayPrediction(gameState, coordinates, stoneColor)
  );
  return getBestPossiblePlay(playPredictions).coordinates;
  
```

PR #91, opened 2022-03-09
Make UI bearable

### LINKED ISSUE #30: Avoid layout flickering
Changes in the rendered contents cause flickering

### LINKED ISSUE #64: Fix winning path not visible for black players on web app
The default font color is black so when black wins the winning path isn't visible

image

fzaninotto:
## Content

- [x] Use tailwind to make the UI more bearable : font family, colors, layout
- [x] Improve makefile and readme
- [x] Avoid flickering by using only one line of text (fixes #30 )
- [x] Change the winning path color to green, to make it visible even on black backgrounds (fixes #64 )
- [x] Show next move suggestion directly on the board
- [x] Fix automated tests

## Demos

image

hex-improve-web-app-ui

### DIFF
--- a/web-app/src/handlebars/helpers.ts
+++ b/web-app/src/handlebars/helpers.ts
@@ -71,6 +71,22 @@ export function registerHandlebarsHelpers() {
       );
     },
   );
+  handlebars.registerHelper(
+    'isCellSuggested',
+    function (
+      suggestedCell: Coordinates,
+      cellXCoord: number,
+      cellYCoord: number,
+      turn: StoneColor,
+    ) {
+      return (
+        suggestedCell &&
+        suggestedCell.x == cellXCoord &&
+        suggestedCell.y == cellYCoord &&
+        turn
+      );
+    },
+  );
 }
 
 export function unregisterHandlebarsHelpers() {
@@ -79,4 +95,5 @@ export function unregisterHandlebarsHelpers() {
   handlebars.unregisterHelper('cellCanBePlayed');
   handlebars.unregisterHelper('needToWaitForOpponent');
   handlebars.unregisterHelper('isCellInWinningPath');
+  handlebars.unregisterHelper('isCellSuggested');
 }
--- a/web-app/views/game.hbs
+++ b/web-app/views/game.hbs
@@ -10,7 +10,7 @@
   </head>
   <body>
 
-    <div class="gameboard">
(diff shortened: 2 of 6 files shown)


web-app/views/game.hbs
```diff
@@ -6,31 +6,25 @@
...
         {{#areEqual game.state.winner null}}
           {{#areEqual readyToPlay true}}
-            <p>
-              Current player:
-              <span class="current-player" data-player="{{game.state.turn}}">{{game.state.turn}}</span>
-            </p>
+            {{#areEqual currentPlayerTurnToPlay true}}
+            <p><span class="current-player" data-player="{{game.state.turn}}">{{game.state.turn}}</span>, your move!</p>
+            {{/areEqual}}
             {{#areEqual currentPlayerTurnToPlay false}}
```
slax57: Suggestion:

```suggestion
            {{else}}
```

(I believe this works event with a custom Handlebars helper 🙂 )

web-app/views/game.hbs
```diff
@@ -6,31 +6,25 @@
...
-              <p>Waiting for other player to make their move...</p>
+              <p>Waiting <span class="current-player" data-player="{{game.state.turn}}">{{game.state.turn}}</span> to make their move...</p>
             {{/areEqual}}
           {{else}}
             <p>Awaiting player 2...</p>
             <p>
               Send <a href="/games/{{game.id}}/join">this link</a> for second player to join
             </p>
```
slax57: Because of these lines, we still have the flickering problem at the beginning of the game. I'll look into that as soon as I can 🙂

slax57: Since it's only at the beginning of the game, it's not that big of a deal. I believed we can postpone this...

atilbian:
Nice work everyone! What do you think about switching to the same colors as in the mobile app ? And I think if the flickering is fixed we can now increase the refresh rate of the gamestate too ! What do you think about that ?

web-app/src/handlebars/helpers.ts
```diff
@@ -71,6 +71,22 @@ export function registerHandlebarsHelpers() {
       );
     },
   );
+  handlebars.registerHelper(
+    'isCellSuggested',
+    function (
+      suggestedCell: Coordinates,
+      cellXCoord: number,
+      cellYCoord: number,
+      turn: StoneColor,
```
atilbian, line 80: Shouldn't the server send an hint only when it's the player's turn ? I really don't have an answer to this so it's ok for me if we don't touch it for now.

PR #93, opened 2022-03-09
Add solo mode where you can play against the AI

slax57:
## Content

- [x] Add an home page button to select the AI mode
- [x] Update game entity to add a isSoloMode
- [x] Update game service funcs to set a game as ready when isSoloMode is true
- [x] Update create game endpoint with a new "soloMode" param
- [x] Update update game endpoint to start the bot play async when soloMode is on
- [x] Add a game service func to make the bot play the suggested move
- [x] ~Include one move to win detection in suggested move function + 2 lines to drop in `getBestPossiblePlay()` - CANCELLED~
- [x] Rebase onto #91 

Related to: https://trello.com/c/xayGxmfu/16-2-as-norbert-i-want-to-be-able-to-be-able-to-play-against-a-bot

## Demo
hex-solo-play

### DIFF
--- a/web-app/public/css/index.css
+++ b/web-app/public/css/index.css
@@ -6,3 +6,66 @@ body {
 form label, form input {
     margin: 10px;
 }
+
+/* The switch - the box around the slider */
+.switch {
+  position: relative;
+  display: inline-block;
+  width: 30px;
+  height: 17px;
+}
+
+/* Hide default HTML checkbox */
+.switch input {
+  opacity: 0;
+  width: 0;
+  height: 0;
+}
+
+/* The slider */
+.slider {
+  position: absolute;
+  cursor: pointer;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: #ccc;
+  -webkit-transition: .4s;
+  transition: .4s;
+}
+
+.slider:before {
+  position: absolute;
+  content: "";
+  height: 13px;
+  width: 13px;
+  left: 2px;
+  bottom: 2px;
+  background-color: white;
+  -webkit-transition: .4s;
+  transition: .4s;
+}
+
+input:checked + .slider {
+  background-color: #2196F3;
+}
+
+input:focus + .slider {
+  box-shadow: 0 0 1px #2196F3;
+}
+
+input:checked + .slider:before {
+  -webkit-transform: translateX(13px);
+  -ms-transform: translateX(13px);
+  transform: translateX(13px);
+}
+
+/* Rounded sliders */
+.slider.round {
+  border-radius: 17px;
+}
+
+.slider.round:before {
+  border-radius: 50%;
+}
--- a/web-app/src/games/games.controller.ts
+++ b/web-app/src/games/games.controller.ts
@@ -26,16 +26,21 @@ export class GamesController {
   @Post('')
   @Redirect('/')
   async createNewGame(
-    @Body() gameParams: { size?: number; fromFile?: boolean },
+    @Body()
+    gameParams: { size?: number; fromFile?: boolean; soloMode?: string },
     @Req() req: Request,
   ) {
     let newGame: Game;
     if (gameParams.fromFile) {
-      newGame = await this.gameService.createNewGameFromFile(req.sessionID);
+      newGame = await this.gameService.createNewGameFromFile(
+        req.sessionID,
+        gameParams.soloMode === 'on',
+      );
     } else {
       newGame = await this.gameService.createNewGame(
         gameParams.size,
         req.sessionID,
+        gameParams.soloMode === 'on',
       );
     }
     return { url: `/games/${newGame.id}` };
@@ -70,7 +75,7 @@ export class GamesController {
     @Req() req: Request,
   ): Promise<GameAndDisplayStatus> {
     let game = await this.gameService.findGameById(id);
-    game = await this.gameService.updateGameState(game, gameParams);
+    game = await this.gameService.applyMove(game, gameParams);
     return this.gameService.getGameAndDisplayStatus(game, req.sessionID);
   }
 }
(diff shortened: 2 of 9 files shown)


web-app/src/games/games.service.ts
```diff
@@ -39,6 +39,18 @@ export class GamesService {
...
   }
 
+  async applyMove(game: Game, coordinates: Coordinates): Promise<Game> {
+    const updatedGame = await this.updateGameState(game, coordinates);
+    if (!updatedGame.state.winner && updatedGame.soloMode) {
+      const nextMove = getNextMoveHint(
+        updatedGame.state,
+        'black',
+      ).suggestedNextMove;
+      this.updateGameState(updatedGame, nextMove);
```
slax57: Should extract this to clarify it's meant to handle the bot turn

web-app/public/css/index.css
```diff
@@ -5,4 +5,67 @@ body {
...
+}
+
+/* Rounded sliders */
+.slider.round {
+  border-radius: 17px;
+}
+
+.slider.round:before {
+  border-radius: 50%;
+}
```
atilbian, line 71: Can be replaced by some tailwind classes : https://tailwind-elements.com/docs/standard/forms/switch/

web-app/src/games/games.controller.ts
```diff
@@ -26,16 +26,21 @@ export class GamesController {
...
+    @Body()
+    gameParams: { size?: number; fromFile?: boolean; soloMode?: string },
     @Req() req: Request,
   ) {
     let newGame: Game;
     if (gameParams.fromFile) {
-      newGame = await this.gameService.createNewGameFromFile(req.sessionID);
+      newGame = await this.gameService.createNewGameFromFile(
+        req.sessionID,
+        gameParams.soloMode === 'on',
```
atilbian: Why not use a boolean instead? Is it to allow to pass ia level later ?

web-app/src/games/games.controller.ts
```diff
@@ -26,16 +26,21 @@ export class GamesController {
...
+      );
     } else {
       newGame = await this.gameService.createNewGame(
         gameParams.size,
         req.sessionID,
+        gameParams.soloMode === 'on',
```
atilbian: If we keep the string value we can at least calculate it only one time at the top of the function and use the variable elsewhere.

web-app/src/games/games.subscriber.ts
```diff
@@ -26,7 +26,11 @@ export class GamesSubscriber implements EntitySubscriberInterface<Game> {
   }
 
   afterUpdate(event: UpdateEvent<Game>): void | Promise<any> {
-    if (event.updatedRelations.some((x) => x.propertyName == 'player2')) {
+    if (
+      event.updatedRelations.some((x) => x.propertyName == 'player2') &&
+      event.entity.player2 &&
+      event.entity.player2.id
```
atilbian, line 32: I would have done it this way but I don't know which one is better :
``` 
event.updatedRelations.some((x) => x.propertyName == 'player2') &&
     !event.entity.soloMode
```

web-app/views/game.hbs
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta charset="utf-8" {{#needToWaitForOpponent readyToPlay currentPlayerTurnToPlay}}http-equiv="refresh" content="5"
+    <meta charset="utf-8" {{#needToWaitForOpponent readyToPlay currentPlayerTurnToPlay}}http-equiv="refresh" content="3"
```
atilbian: As a memo, don't forget to set it lower(1 as in the mobile app?)

web-app/views/index.hbs
```diff
@@ -13,6 +13,12 @@
   <form action="/games" method="post">
     <label for="size">Board size:</label><br>
     <input type="number" id="size" name="size" value="19" data-cy="size"><br>
+    <label for="soloMode">Play against a bot</label>
+    <label class="switch" for="soloMode">
+      <input type="checkbox" id="soloMode" name="soloMode">
```
atilbian: We can add a value prop here set to "true" instead of the default value !

web-app/src/games/games.service.ts
```diff
@@ -106,18 +136,21 @@ export class GamesService {
...
-      readyToPlay: this.haveTwoPlayersJoinedIn(game),
-      currentPlayerTurnToPlay:
+      readyToPlay: this.haveTwoPlayersJoinedIn(game) || game.soloMode,
+      currentPlayerTurnToPlay: !!(
         (game.state.turn === 'white' && playerName === game.player1.username) ||
-        (game.state.turn === 'black' && playerName === game.player2.username),
+        (game.state.turn === 'black' &&
+          game.player2 &&
+          playerName === game.player2.username)
+      ),
```
atilbian: I think it becomes harder to read it and could benefit from being extracted into it's own well named function instead, e.g isItMyTurnToPlay(askingPlayer, game). Ok it's not very well named but i don't have a better idea for now ^^

web-app/src/games/games.controller.ts
atilbian: I have my answer from https://github.com/marmelab/Hex/pull/93#discussion_r823513758 and i let you choose if you want to switch to a bool val or keep the default val.

web-app/src/games/games.subscriber.ts
slax57: I prefer my way of doing it since the link between the presence of `event.entity.player2.id` and the ability to perform `this.userService.incrementNbGames(event.entity.player2.id);` is obvious , whereas the link with `event.entity.soloMode` is not guaranteed to stay true in the future

web-app/views/game.hbs
slax57: I'll try with 1 and see how it looks 🙂

web-app/public/css/index.css
slax57: Introduced too much side work. I'll leave the current CSS and won't use tailwind on this page for now.

web-app/src/games/games.subscriber.ts
atilbian: Ok it's good for me !

PR #94, opened 2022-03-09
Use minimax to make better suggestions

atilbian:
- [X] Implem the minimax algo
- [X] ~Add additional tests for complex cases to validate that it works as expected~ For the next PR
- [X] ~Use the binary representation for board~ For the next PR
- [X] ~Replace the use of the previous getNextPlaySuggestion by the minimax one in tests(at depth = 1)~ For the next PR

## Minimax
image


slax57 (approved):
Great job! 🎉

